PR #233, opened 2023-06-21
feat: add support for Procedure and ServiceRequest resources

mikix:
Two new tasks:
- procedure
- servicerequest

i2b2 support for these has not been added yet, that can happen separately.

### Checklist


cumulus_etl/formats/batched_files.py
```diff
@@ -47,9 +47,9 @@ def __init__(self, *args, **kwargs) -> None:
             self.root.rm(parent_dir, recursive=True)
         except FileNotFoundError:
             pass
-        self.root.makedirs(parent_dir)
```
mikix, line 50: Unrelated, but I found that we were making folders for resources that didn't get written when running i2b2, because we made each task's folder early. Just moving that to when we write a batch out, so that we don't make so many spurious folders.

dogversioning (approved):

PR #234, opened 2023-06-22
feat: when writing delta lake tables, include a shallow FHIR schema

mikix:
This helps downstream SQL not have to worry about whether the incoming data has a particular optional FHIR field (which before, would mean that the Athena schema didn't define it, which would yield a SQL error).

Now, all the R4 FHIR fields will be represented at least at the toplevel, and SQL can rest a bit easier.

### Checklist

### DIFF
--- a/cumulus_etl/chart_review/cli.py
+++ b/cumulus_etl/chart_review/cli.py
@@ -8,7 +8,7 @@ from collections.abc import Collection
 import ctakesclient
 from ctakesclient.typesystem import Polarity
 
-from cumulus_etl import cli_utils, common, deid, errors, fhir_client, fhir_common, loaders, nlp, store
+from cumulus_etl import cli_utils, common, deid, errors, fhir, loaders, nlp, store
 from cumulus_etl.chart_review import downloader, selector
 from cumulus_etl.chart_review.labelstudio import LabelStudioClient, LabelStudioNote
 
@@ -29,7 +29,7 @@ def init_checks(args: argparse.Namespace):
 
 
 async def gather_docrefs(
-    client: fhir_client.FhirClient, root_input: store.Root, root_phi: store.Root, args: argparse.Namespace
+    client: fhir.FhirClient, root_input: store.Root, root_phi: store.Root, args: argparse.Namespace
 ) -> loaders.Directory:
     """Selects and downloads just the docrefs we need to an export folder."""
     common.print_header("Gathering documents...")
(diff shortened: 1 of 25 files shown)


cumulus_etl/chart_review/cli.py
```diff
@@ -8,7 +8,7 @@
 import ctakesclient
 from ctakesclient.typesystem import Polarity
 
-from cumulus_etl import cli_utils, common, deid, errors, fhir_client, fhir_common, loaders, nlp, store
+from cumulus_etl import cli_utils, common, deid, errors, fhir, loaders, nlp, store
```
mikix, line 11: You're going to see a lot of light refactoring. I was adding a new "FHIR spec"-like bit of code, so I took the opportunity to gather the FHIR code into one subdirectory.

`fhir_client` -> `fhir`
`fhir_common` -> `fhir`
(new) `fhir_schemas` -> `fhir`

And I moved the general purpose `FatalError` out of `fhir_client` and into `errors`.

pyproject.toml
```diff
@@ -13,14 +13,15 @@ requires-python = ">= 3.10"
 dependencies = [
     "ctakesclient >= 3, < 4",
     "delta-spark >= 2.3, < 3",
+    "fhirclient < 5",
```
mikix, line 16: I know we just got rid of `fhirclient`, but it was the quickest/easiest way I could find to get python-friendly flattened versions of the FHIR spec.

pyproject.toml
```diff
@@ -13,14 +13,15 @@ requires-python = ">= 3.10"
...
     "httpx < 1",
     "inscriptis < 3",
     "jwcrypto < 2",
     "label-studio-sdk < 1",
     "oracledb < 2",
     "pandas < 3",
     "philter-lite < 1",
-    "pyarrow < 12",
+    "pyarrow < 13",
```
mikix, line 24: Not necessary, I just noticed we were a version back while I was playing with data transforms.

cumulus_etl/formats/deltalake.py
```diff
@@ -165,3 +157,45 @@ def _configure_fs(root: store.Root, spark: pyspark.sql.SparkSession):
         region_name = fsspec_options.get("client_kwargs", {}).get("region_name")
         if region_name:
             spark.conf.set("fs.s3a.endpoint.region", region_name)
+
+    @contextlib.contextmanager
+    def pandas_to_spark_with_schema(self, dataframe: pandas.DataFrame) -> pyspark.sql.DataFrame:
+        """Transforms a pandas DF to a spark DF with a full FHIR schema included"""
+        # This method solves two problems:
+        # 1. Pandas schemas are very loosey-goosey (and don't really take nested data into account), so simply
```
dogversioning, line 165: such accurate spelling of loosey-goosey!

cumulus_etl/chart_review/cli.py
dogversioning: this generally feels like a very good move to me.

dogversioning (approved):

mikix:
Tested this on some BCH delta lakes and it seemed to work & fix our issues with the `core` library. Merging away.

PR #236, opened 2023-06-23
fix: avoid schema-mismatch errors for int columns that have nulls

mikix:
Pandas converts integer columns with nulls to float columns by default. This confused our new schema merging feature.

So now we ask Pandas to use more modern column types (ones that can handle nulls better). This fixes the schema issue.

BUT it might introduce incompatibilities with existing tables. If so, the easiest fix is to blow away the table and recreate it.

### Checklist

### DIFF
--- a/cumulus_etl/loaders/i2b2/extract.py
+++ b/cumulus_etl/loaders/i2b2/extract.py
@@ -19,10 +19,10 @@ def extract_csv(path_csv: str, batch_size: int) -> Iterator[dict]:
         with pandas.read_csv(path_csv, dtype=str, na_filter=False, chunksize=batch_size) as reader:
             print(f"Reading csv {path_csv}...")
             for chunk in reader:
-                print(f"  Read {count:,} entries...")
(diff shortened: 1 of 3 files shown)


cumulus_etl/loaders/i2b2/extract.py
```diff
@@ -19,10 +19,10 @@ def extract_csv(path_csv: str, batch_size: int) -> Iterator[dict]:
         with pandas.read_csv(path_csv, dtype=str, na_filter=False, chunksize=batch_size) as reader:
             print(f"Reading csv {path_csv}...")
             for chunk in reader:
-                print(f"  Read {count:,} entries...")
                 for _, row in chunk.iterrows():
                     yield dict(row)
                 count += batch_size
+                print(f"  Read {count:,} entries...")
```
mikix, line 25: Unrelated, but I was dealing with some small-scope batches recently and noticed that this was printing `read 0 entries` and got confused.

dogversioning (approved):

PR #237, opened 2023-06-27
feat: download Medication resources as we process MedicationRequests

mikix:
Since many EHRs (Epic & Cerner at least) don't allow bulk downloading Medication resources (it's not a patient-specific resource), this commit adds the ability to download any referenced Medications inside MedicationRequest.

This adds the ability for any task to have multiple output tables and to provide multiple streams of output entries as it reads the input resources.

Only the MedicationRequest task uses this ability so far.

Other related and unrelated changes described below.

Tasks:
- Split tasks.py up into a submodule, it was getting too big.
- Support multiple output tables for tasks and re-arrange how a task subclass defines output targets. This also means we support task.read_entries() with multiple simultaneous data streams. For example, MedicationRequest will often provide itself plus a Medication as a tuple result from read_entries() and the task runner will know they are separate tables.
- Any task that runs, even if it doesn't have any output, will now write out an empty dataframe -- this ensures that the table exists with a schema, to simplify downstream SQL (a sibling effort to the recent addition of schemas at all)

UX:
- Stop printing json summaries at end of ETL run - the task outputs themselves should be sufficient and the json summaries were very technical, not user-oriented.
- Fix i2b2 entry-reading counts (again)

Bulk export:
- Don't check for token_endpoint_auth_signing_alg_values_supported declarations by the server. Epic doesn't provide it, but supports SMART jwks anyway. This fixes bulk exporting with Epic.

Convert subcommand:
- Write a schema too, this got missed with the recent schema PR.
- Read ndjson ourselves, which gives us better type detection (no more birthDate being parsed as an int instead of a string)

### Checklist


cumulus_etl/etl/tasks/base.py
```diff
@@ -0,0 +1,254 @@
+"""ETL tasks"""
```
mikix, line 1: I split up the single tasks.py into a submodule, because it was getting too large.

Unfortunately, along with that refactor, I'm also changing some of the code which makes reviewing a little hard. I could have maybe split them into different commits or PRs to be kind, but I was lazy and here we are.

All the changes here are around supporting multiple output tables (so some interesting changes to the batching iteration utility methods to handle tuples of simultaneous output streams, some changes to the base class EtlTask run support to hand those output streams to multiple Format instances, etc).

The other big change in this folder is the rewrite of MedicationRequestTask to use that new support and download Medications on the fly. So if you want to focus on that rather than the refactor, go look at just that class.

cumulus_etl/etl/tasks/basic_tasks.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # Keep a cache of medication IDs that we've already downloaded.
+        # If this ends up growing too large in practice, we can wipe it during a call to table_batch_cleanup().
+        # But let's try initially with keeping it around for the whole task.
+        self.medication_ids = set()
+
+    def scrub_medication(self, medication: dict) -> bool:
```
mikix, line 53: This method is maybe something to look at with side-eye. I'm doing some manual de-identification (rather than via the MS tool).

I figured this made sense for technical reasons (MS tool only operates on a pile of ndjson on disk, and normally only runs once at the beginning of ETL) and it-isn't-patient-data reasons (these resources have no connection to patients at all).

But still... I'm open to feedback. I *could* do it via MS tool by writing all these medications to disk and then running it on them, then reading that back in.

Here's the spec for this resource, FYI: https://www.hl7.org/fhir/R4/medication.html

dogversioning (approved):
This is mostly based on looking at the medication handler - I took a quick pass through the rest of the task refactor and it seems reasonable enough, but it's a lot and i didn't go very deep on it.

cumulus_etl/etl/tasks/basic_tasks.py
mikix: For posterity: I checked with Jamie and Andy, they think this is OK.

But also, in further chats with you, we may want to allow ingesting Medications like normal resources anyway -- for (A) future EHRs that do let you bulk export them and (B) cases like Synthea where you have pre-generated resources.

In those cases, we may want to switch to an MS config for both flows and write out all Medications to a local ndjson and re-ingest it.

But for now, let's go with this and manually deal with Synthea tables.

PR #238, opened 2023-06-28
added desc lookup for I2B2

dogversioning:
This PR makes the following changes:
- Adds a code dict for display text lookup
- Inserts data during the i2b2 transform step

### Checklist


cumulus_etl/loaders/i2b2/icd.json
dogversioning, line 1: This could live someplace else if you want, but this seemed ok based on how we are scoping this?

cumulus_etl/loaders/i2b2/transform.py
```diff
@@ -15,6 +18,8 @@
 ###############################################################################
 
 
+ICD_CODE_DICT = json.load(open(Path(Path(__file__).resolve().parent,'icd.json')))
```
dogversioning, line 21: It seemed silly to me to load in a 10MB json dict a million times - again, this could be done someplace else if you prefer.

cumulus_etl/loaders/i2b2/transform.py
```diff
@@ -347,4 +352,6 @@ def make_concept(code: str, system: str | None, display: str = None) -> dict:
     coding = {"code": code, "system": system}
     if display is not None:
         coding["display"] = display
+    elif system in ICD_CODE_DICT:
+        coding["display"] = ICD_CODE_DICT[system].get(code, 'Display not found')
```
mikix, line 356: This seems intentional, but is `Display not found` the text we want to show up in graphs? Vs leaving as `None` and letting the library or dashboard put a string in (my gut instinct)? Or even something vaguer/less-techy like `Unknown`?

cumulus_etl/loaders/i2b2/transform.py
mikix: I'm a little sensitive to memory usage - we do a lot of stuff to not keep data in memory if we don't have to. Can we load this in a way that would allow it to be GC'd when done with it?

Obviously don't load it a million times. But load it once per i2b2 load, then null it out. The current code structure doesn't have an obvious place to do that. Do you have thoughts there?

You could kind of hack something for that by adding like init and cleanup methods to `transform.py` that maybe `_load_all_with_extractors` could call? Doesn't feel elegant, but might be easiest.

Another option (which feels more elegant and just as easy?) is to do something like this in `I2b2Loader`:
```
from functools import partial
...
        if "Condition" in resources:
            code_dict = load_big_json_file()
            self._loop(
                conditions(),
                partial(transform.to_fhir_condition, code_dict),
                os.path.join(tmpdir.name, "Condition.ndjson"),
            )
``` 
And change to_fhir_condition to accept a code dict as the first param.

At the end of the day, 10MB ain't the worst. But still.

cumulus_etl/loaders/i2b2/icd.json
mikix: I like it here for now. We can move when/if it is used more broadly.

cumulus_etl/loaders/i2b2/transform.py
dogversioning: i can live with Unknown. 

There's a 'no one wants to to own this piece' kind of argument going on here - we really want this to come from the data source directly? and if it weren't this pre-FHIR legacy thing we'd be going back to them? But an 'Unknown' string :could: be easily replaced in either of the other layers if we needed it to.

cumulus_etl/loaders/i2b2/transform.py
dogversioning: yeah i was originally thinking about kwarg passing, this was just conceptually faster for something with some time pressure - i could go up the call chain and add the dict loader in at that entry point and be explicit about it.

currying functions hurts my brain. if you prefer that i can do it, but that's not normally something i would even think to try in a vacuum.

mikix: Hah, I thought you loved functional programming and would be all about it. :smile: 

I don't particularly care about `partial`, but I thought it might avoid a larger refunction of that whole `_loop` setup that is otherwise very generic. (though I guess you could just pass `**kwarg` down through, so :shrug: easy either way, as I type this.)

If this is urgent, you can land this and clean up later - 10MB will not break us. But the kwarg approach seems good to me as a cleanup (now or later). Will approve to get out of your way.

mikix (approved):

cumulus_etl/loaders/i2b2/transform.py
dogversioning: i think that functional programming is very good for memory usage and once you get past map/reduce i get very fuzzy on low level details, especially since it's been a year or two since i've had to do any, and my brain is a leaky bucket.

PR #241, opened 2023-07-05
fix: allow contained Medication resources through

mikix:
- Start anonymizing contained "id" fields (that's an oversight that should have already happened, but contained resources are not widely used and I don't expect this to require much backfilling)
- Add an MS tool config for Medication, so that any MedicationRequests that contain a Medication will allow its interesting values through.

### Checklist

### DIFF
--- a/tests/data/mstool/input/MedicationRequest.2.json
+++ b/tests/data/mstool/input/MedicationRequest.2.json
@@ -1,0 +1,21 @@
+{
+  "resourceType": "MedicationRequest",
+  "id": "2-contained",
+  "contained": [
+    {
+      "resourceType": "Medication",
+      "id": "contained",
+      "status": "active",
+      "code": {
+        "coding": [
+          {
+            "system": "http://hl7.org/fhir/sid/ndc",
+            "code": "0069-2587-10",
+            "display": "Vancomycin Hydrochloride (VANCOMYCIN HYDROCHLORIDE)"
+          }
+        ]
+      }
+    }
+  ],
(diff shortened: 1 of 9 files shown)


tests/data/mstool/input/MedicationRequest.2.json
```diff
@@ -0,0 +1,21 @@
+{
+  "resourceType": "MedicationRequest",
+  "id": "2-contained",
+  "contained": [
```
dogversioning, line 4: this may be over paranoid, but could this test data benefit from an additional non-medication resource?

dogversioning (approved):

tests/data/mstool/input/MedicationRequest.2.json
mikix: I test "unknown contained resource" handling elsewhere - you can see in this PR, I'm actually switching the "unknown resource" in Condition.1.json from Medication to CarePlan, because now Medication is known.

But you are trying to catch an accidental rule like `keep MedicationRequest.contained.*`?

dogversioning: if that's what that's doing, then i think my main concern is addressed - my main thought was a pedantic 'do we need to test a resource which has both an expected and an unexpected resource at the same time', but as long as you're hitting both individually i'm ok.

mikix: Condition.1 actually does test "both expected and unexpected at same time" :+1:

PR #243, opened 2023-07-07
fix: make writes to remote locations atomic

mikix:
Also, reduce the number of times that the codebook writes itself out (i.e. stop writing the main codebook every time we write out the mappings)

### Checklist

### DIFF
--- a/cumulus_etl/common.py
+++ b/cumulus_etl/common.py
@@ -7,10 +7,9 @@ import json
 import logging
 import re
 from collections.abc import Iterator
-from typing import Any
-from urllib.parse import urlparse
+from typing import Any, TextIO
 
-import fsspec
+from cumulus_etl import store
 
 
 ###############################################################################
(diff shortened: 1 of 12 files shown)


cumulus_etl/common.py
```diff
@@ -32,49 +31,13 @@ def ls_resources(root, resource: str) -> list[str]:
 #
 ###############################################################################
 
-_user_fs_options = {}  # don't access this directly, use get_fs_options()
```
mikix, line 35: This bit of business with tracking use-filesystem options (like --s3-region) got moved to `store` because that flips the relationship between them - now `common` uses `store` and can make use of `store.Root`. (and it makes more sense for these FS options to live there)

cumulus_etl/common.py
```diff
@@ -32,49 +31,13 @@ def ls_resources(root, resource: str) -> list[str]:
...
-    protocol = parsed.scheme or "file"  # assume local if no obvious scheme
-
-    # We pass auto_mkdir because on some backends (like S3), we may not have permissions that fsspec might want,
-    # like CreateBucket. We elsewhere call Root.makedirs as needed.
-    return fsspec.open(path, mode, encoding="utf8", auto_mkdir=False, **get_fs_options(protocol))
+@contextlib.contextmanager
+def open_file(path: str, mode: str) -> TextIO:
+    """A version of open() that handles atomic file access across many remote filesystems (like S3)"""
+    root = store.Root(path)
+    with root.fs.transaction:
```
mikix, line 39: :boom: this is the main new line in the PR that does all the magic

cumulus_etl/common.py
```diff
@@ -32,49 +31,13 @@ def ls_resources(root, resource: str) -> list[str]:
...
-
-def open_file(path: str, mode: str):
-    """A version of open() that handles remote access, like to S3"""
-    # Grab protocol if present
-    parsed = urlparse(path)
-    protocol = parsed.scheme or "file"  # assume local if no obvious scheme
-
-    # We pass auto_mkdir because on some backends (like S3), we may not have permissions that fsspec might want,
-    # like CreateBucket. We elsewhere call Root.makedirs as needed.
-    return fsspec.open(path, mode, encoding="utf8", auto_mkdir=False, **get_fs_options(protocol))
```
mikix, line 77: auto_mkdir logic here does not need to be replicated - it's on by default only for the main entry points `fsspec.open` and `fsspec.open_files` -- actual filesystems have it off by default. :shrug:

cumulus_etl/deid/codebook.py
```diff
@@ -242,16 +242,16 @@ def save(self, codebook_dir: str) -> bool:
...
         saved = False
 
-        if self.modified:
-            logging.info("Saving codebook to: %s", codebook_dir)
-
+        if self.settings_modified:
             codebook_path = os.path.join(codebook_dir, "codebook.json")
             common.write_json(codebook_path, self.settings)
+            self.settings_modified = False
+            saved = True
```
mikix, line 249: Before this, any modification of the cached Patient/Encounter mappings would write out the codebook settings all over again, and these days the codebook settings never change.

(In BCH's legacy i2b2 environment, this meant writing out 1G every time we had a new mapping... :frowning: )

cumulus_etl/etl/tasks/base.py
```diff
@@ -145,7 +145,8 @@ async def _write_tables_in_batches(self, entries: EntryIterator) -> None:
                 self.table_batch_cleanup(table_index, batch_index)
 
                 print(f"  {summary.success:,} processed for {formatter.dbname}")
-                batch_index += 1
+
+            batch_index += 1
```
mikix, line 149: Unrelated, but noticed that this counter had the wrong indenting. This didn't actually matter, unless someone was doing ndjson exports of medicationrequests.

cumulus_etl/common.py
```diff
@@ -32,49 +31,13 @@ def ls_resources(root, resource: str) -> list[str]:
...
+@contextlib.contextmanager
+def open_file(path: str, mode: str) -> TextIO:
+    """A version of open() that handles atomic file access across many remote filesystems (like S3)"""
```
dogversioning, line 37: - how would you feel about calling this `atomic_open_file` or `open_remote_file`?
- re the docstring, i believe you, but it's not apparent from me looking at the function why this provides better atomicity.works.

dogversioning (approved):
minor semantic suggestion, approving to get out of the way.

cumulus_etl/common.py
mikix: > how would you feel about calling this atomic_open_file or open_remote_file?

Hmm, `open_remote_file` is misleading a bit, because it also handles local atomicity. I know the docstring calls out remote, but that's because that's harder and more impressive to offer. :smile: 

I can remove the word `remote` there though to avoid confusion.

As for `atomic_open_file` -- I'm a little resistant but could be convinced. (A) I'd probs want to mark it with an underscore then because that's a bit of a mouthful for a public method. (B) emphasizing the atomicity like that was not my first instinct because really, this is just fixing `open_file` to work like the ETL wants (and most sane high-level-I/O folks would expect) -- we never want non-atomic writes in any part of our ETL, this is just how the ETL _does_ I/O, so there's not like two versions of this, one with an atomic prefix and one without.

> re the docstring, i believe you, but it's not apparent from me looking at the function why this provides better atomicity/works.

I can add a comment next to the transaction line - but `fsspec` by default is not atomic - it will write out what it has done so far upon an error. So the `with transaction` line in this method is how fsspec offers atomicity - it will no longer commit the file upon an error, but instead abandon it.

(For local filesystems, it writes to temp and moves the file into place when done. For S3, it always uploads files as non-visible multipart chunks anyway, and either commits them to the final place on success or abandons the chunks on failure. Without a transaction in place, fsspec commits by default even on failure)

mikix: Cleaned up docstring and added a comment near the transaction line

mikix: Hmm, actually the ETL does sometimes do partial writes - to local ndjson log/download files  -- but that's already abstracted separately.

mikix: Hmm, maybe I'm being resistant for no reason - I'll rename it `_atomic_open()`

PR #246, opened 2023-07-10
Return of the Atomic Writes

mikix:
This PR contains 3 commits:
- Old PR #243 (reviewed, reverted)
- Old PR #244 (not reviewed, reverted)
- New commit that fixes the remaining issue.

This new commit has the actual fix to a bug introduced in #243:
```
-        return root.fs.open(path, mode=mode, encoding="utf8")
+        with root.fs.open(path, mode=mode, encoding="utf8") as file:
+            yield file
```

The original commit changed `open_file` (a method that returns a file) to `_atomic_open` (a context manager that returns a file), inadvertently dropping the bit where we were actually using the file as a context manager itself, which means the file was not being closed and could have lingering data not committed to disk.

Easy to miss, hard to find.

I've also updated our testing to catch the several error cases that arise when you forget to close, so they are a bit paranoid now (confirm that small and large writes, to local and s3, using text or json all work correctly).

And I've done some manual testing to confirm everything seems to work.

### Checklist

### DIFF
--- a/tests/s3mock.py
+++ b/tests/s3mock.py
@@ -1,101 +1,74 @@
 """
 Support for mocking out an S3 server during a test.
-
-ALso, patch aiobotocore (used by pandas) to work with moto
-See https://github.com/aio-libs/aiobotocore/issues/755
 """
 
 import os
-import unittest
-from collections.abc import Callable
-from unittest.mock import MagicMock, patch
 
-import aiobotocore.awsrequest
-import aiobotocore.endpoint
-import aiohttp
-import aiohttp.client_reqrep
-import aiohttp.typedefs
-import botocore.awsrequest
-import botocore.model
-from moto import mock_s3
+import moto
+import s3fs
+from moto.server import ThreadedMotoServer
+
+from cumulus_etl import store
+from tests import utils
 
 
-class S3Mixin(unittest.TestCase):
+class S3Mixin(utils.AsyncTestCase):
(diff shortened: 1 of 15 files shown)


tests/s3mock.py
mikix, line 1: This file saw a lot of changes in the most recent commit - trying to use `moto` more heavily was making me crazy with even further conflicts with aiobotocore. After some reading, I found that using moto as a spun up server avoids that conflict, so I rewrote how this mixin works, at the cost of some complexity (but drops all that goofy patching, so I think we win on net-complexity).

dogversioning: yeah i buy it.

random food for thought - any value in setting up a conftest.py for fixtures like this?

dogversioning (approved):
Only really looked at the changes in the most recent commit, since it supersedes the unreviewed PR and I already looked at the first one.

tests/s3mock.py
mikix: Let's chat about that some time - I'm no stranger to conftest.py (I use one in ctakesclient to autoskip some tests if you don't have a running NLP server), but my suspicion is that putting fixtures there is mostly useful if you're making pytest-style tests (toplevel methods & fixture-decorators) vs unittest-style tests (classes & mixins & decorators).

So far it seems your brain and mine just have different value sets there. But happy to learn how conftest.py could make my life better.

PR #247, opened 2023-07-13
fix: make sure that no inferred schemas ever hit a delta lake

mikix:
This commit fixes a couple oversights with inferred schemas:
- The type might clash with the wide schema we have per-spec, even though the types are really compatible (inferred int for a float field does happen)
- The wrong type might get through unnoticed if it is a deep field that our per-spec schema didn't catch.

Here's a summary of changes to make that happen:
- Drop all use of pandas. It's too loose with the types. Instead, switch to pyarrow which is used under the covers anyway.
- Add schema earlier in the process (at Task batching time, not at Formatter writing time). This means that all formatters get the same nice schema.

~~:rotating_light: I am worried that some of the things here will increase memory use (i.e. create multiple copies of batches). Like how I create a per-column version of the data, to get pyarrow to detect the shape of the data. I did a quick and dirty check of peak memory before and after this PR and we went from 11x batch size to 20x batch size. So like, twice the memory? But neither situation is good. I'd hoped we'd be like, 2x batch size at worst. But maybe my test is bad? I think I need to look into peak memory use and keep a unit test to avoid regressions. But for now, I think higher memory use but safe schemas, is better than not... :rotating_light:~~

### Checklist


cumulus_etl/common.py
```diff
@@ -37,10 +39,14 @@ def _atomic_open(path: str, mode: str) -> TextIO:
...
-    # fsspec is atomic per-transaction -- if an error occurs inside the transaction, partial writes will be discarded
-    with root.fs.transaction:
-        with root.fs.open(path, mode=mode, encoding="utf8") as file:
-            yield file
+    with contextlib.ExitStack() as stack:
+        if "w" in mode:
+            # fsspec is atomic per-transaction.
+            # If an error occurs inside the transaction, partial writes will be discarded.
+            # But we only want a transaction if we're writing - read transactions may error out
+            stack.enter_context(root.fs.transaction)
```
mikix, line 47: I don't know why we weren't hitting this before - I was getting test errors on I think fake memory/buffer input files that didn't expect certain fsspec semantics - but it makes sense to skip the transaction on read anyway.

cumulus_etl/etl/tasks/base.py
```diff
@@ -173,41 +178,60 @@ def _get_formatter(self, table_index: int) -> formats.Format:
...
+        id_set = set()
+
+        def is_unique(row):
+            nonlocal id_set
+            if row["id"] in id_set:
+                return False
+            id_set.add(row["id"])
+            return True
+
+        return [row for row in batch if is_unique(row)]
```
dogversioning, line 204: this feels like a good memory profiling target to me.

cumulus_etl/fhir/fhir_schemas.py
```diff
@@ -3,16 +3,27 @@
...
-def create_spark_schema_for_resource(resource_type: str) -> pyspark.sql.types.StructType:
+def pyarrow_schema_from_resource_batch(resource_type: str, batch: list[dict]) -> pyarrow.Schema:
+    # Switch batch from a list of rows to a list of columns (which lets pyarrow infer full unionized struct types
+    # across all column values -- it seems to normally only want to infer from the first row if you use any of its
+    # row-oriented APIs, but if you give it columns, it looks at the whole set).
+    column_names = get_all_column_names(resource_type)
+    columns = {col_name: pyarrow.array(row.get(col_name) for row in batch) for col_name in column_names}
+    inferred_schema = pyarrow.struct({name: array.type for name, array in columns.items()})
+
+    return create_pyarrow_schema_for_resource(resource_type, inferred_schema)
```
dogversioning, line 21: is it possible to create a schema from the expected FHIR datatype without inspecting the batch?

cumulus_etl/formats/deltalake.py
```diff
@@ -161,52 +162,11 @@ def _configure_fs(root: store.Root, spark: pyspark.sql.SparkSession):
...
-        # - Our FHIR schema is incomplete and shallow (it skips all nested structs) to avoid infinite recursion issues,
-        #   and we simply merge this incomplete schema in with the actual data schema, which will have full nested
-        #   inferred schemas for exactly the fields it uses. We always write to the delta lake with autoMerge of
-        #   schemas enabled, so incrementally adding fields to existing lakes will be fine.
-        # - Delta Lake does not like columns that have null types nor struct types with no children. So we make sure
-        #   that every column has *some* definition, and that structs have content.
-
+    def pyarrow_to_spark(self, table: pyarrow.Table) -> pyspark.sql.DataFrame | None:
+        """Transforms a pyarrow table to a spark DF"""
+        # There must be a better way?
```
dogversioning, line 167: oh dear, one only hopes.

dogversioning (commented):
just one or two comments of questionable use while you dig into deeper debugging.

cumulus_etl/formats/deltalake.py
mikix: In the grand scheme of things, this is a small oddity - we were writing out pandas to parquet and reading back into sparks before (to get proper schemas by way of pyarrow). And now that we have pyarrow directly, I'm just swapping in pyarrow for pandas.

Searching for how to do direct pyarrow -> pyspark yields people talking about doing this (parquet intermediate) or going to pandas as an intermediate (but that way lies bad schemas, which is why even when we used pandas, we wrote to a parquet).

The native-rust [python bindings](https://github.com/delta-io/delta-rs/blob/main/python/deltalake/table.py) seem to be very pyarrow focused, so that's promising for the future...

cumulus_etl/fhir/fhir_schemas.py
mikix: Yes - and that's basically what we did before (the "wide" version).

But this code is adding the feature of "wide plus any deep fields in the batch" which requires inspecting the batch. The quick and dumb way I did was to use pyarrow's inferring support, but for that to work, it needs to examine columnar data.

Instead, what I could do is write some code that inspects every row/column of the batch and discovers the maximal "shape" of the structs values for those columns, so we know what fields to include in the final schema. Which is basically what pyarrow is doing. But with some memory cost.

cumulus_etl/etl/tasks/base.py
mikix: This costs one extra batch size plus a set of of IDs, and is temporary. It's definitely adding to the problem, but this is (in my mind) an acceptable memory cost since it is reaped back so quickly. That is, my ideal ETL would peak at roughly 2x the batch size in memory.

Whether it's the most efficient, time or memory wise, is certainly an open topic.

cumulus_etl/fhir/fhir_schemas.py
mikix: I ended up writing that code that inspects the shape manually (rather than converting the row-oriented data to column-oriented data). It wasn't so bad. Look for `merge_shape_of_dict()` in `fhir_schemas.py`

cumulus_etl/formats/deltalake.py
mikix: I added a fuller comment to this now that `deltalake.py` is taking in a list of dicts rather than a pyarrow Table. I know of a better way to do this, but I'm leaving it for a future improvement:

```
        # This is the quick and dirty way - write batch to parquet with pyarrow and read it back.
        # But a more direct way would be to convert the pyarrow schema to a pyspark schema and just
        # call self.spark.createDataFrame(batch.rows, schema=pyspark_schema). A future improvement.
```

mikix:
@dogversioning OK I re-wrote portions of this PR to keep the incoming data stream as a simple list of dictionaries for as long as possible. The conversions to and from pyarrow Tables were adding too much memory spikes. So we now only convert data when we need to at the last mile.

Please re-review when you can. But this PR no long makes our memory use worse.

cumulus_etl/fhir/fhir_schemas.py
```diff
@@ -3,78 +3,123 @@
...
     can assume each column is at least defined.
     """
+    # Examine batch to see the full shape of it, in order to detect any deeply nested fields that we want to make sure
+    # to include in the final schema (normally, we go wide but only as deep as we need to)
+    batch_shape = merge_shape_of_dict(None, batch)
+
+    return create_pyarrow_schema_for_resource(resource_type, batch_shape)
+
+
+def merge_shape_of_dict(total: dict | None, item) -> dict:
```
dogversioning, line 31: i'm bouncing off this name a bit - perhaps `get_shape_of_merged_dict`?

does this function need a docstring?

cumulus_etl/formats/batch.py
dogversioning, line 1: for the record, i am pretty sure this is not an issue today - do we want to give this a slightly more descriptive name? like, if we end up with a different batching system for some reason in the future, is this too generic? or do we make that change when we come to it?

dogversioning (approved):

cumulus_etl/fhir/fhir_schemas.py
mikix: Yeah OK for sure a docstring and I'll try to think of a better name - my brain said `merge` because it's a recursive thing that is always merging back changes into the master dict. But that can be clearer / better worded. And yeah a docstring. Good call out.

cumulus_etl/formats/batch.py
mikix: I'll lean defer finding a different name, just because I'm lazy but also I (and the ETL by extension) have been using "batch" as a sort of term of art around "a chunk of input data that winds its way through the ETL, whose size is controlled by `--batch-size`"

Maybe at minimum, this deserves a nice docstring saying at least that.

cumulus_etl/fhir/fhir_schemas.py
mikix: OK done - hopefully this is a little clearer (and I included an example in the docstring)

PR #258, opened 2023-08-01
feat: add cross-task temporary dir to share downloaded notes

mikix:
This commit adds a global temporary directory that can be used to share data between disparate bits of code/tasks.

Specifically, the first user of this is fhir_utils.get_docref_note which now caches any downloaded notes in this temporary directory.

This will make it much more performant to add a second NLP task (it won't have to re-download the notes).

Like all temporary directories managed via context manager, this directory will be cleaned up upon ETL finishing or being interrupted by an error.

It is also created with mode 0700, so only the ETL user can read the files.

### Checklist


cumulus_etl/fhir/fhir_utils.py
```diff
@@ -190,6 +218,7 @@ async def get_docref_note(client: fhir.FhirClient, docref: dict) -> str:
 
     # Strip this "line feed" character that often shows up in notes and is confusing for NLP.
     # Hopefully not many notes are using actual Spanish.
-    note = note and note.replace("¿", " ")
+    note = note.replace("¿", " ")
```
mikix, line 221: Unrelated, but a while back I changed this method to not use `None` notes and instead just raise exceptions on errors. So this `note and` check is unnecessary now and gives the wrong impression that `note` could be `None`.

cumulus_etl/cli.py
```diff
@@ -48,16 +49,20 @@ async def main(argv: list[str]) -> None:
...
         if not subcommand:
             # Add a note about other subcommands we offer, and tell argparse not to wrap our formatting
             parser.formatter_class = argparse.RawDescriptionHelpFormatter
             parser.description += "\n\n" "other commands available:\n" "  chart-review\n" "  convert"
-        await etl.run_etl(parser, argv)
+        run_method = etl.run_etl
+
+    with tempfile.TemporaryDirectory() as tempdir:
+        common.set_global_temp_dir(tempdir)
+        await run_method(parser, argv)
```
mikix, line 65: I set this up here for every subcommand we have. Which might be overkill? Like, this first use of it is only really needed in the ETL. But... it seems harmless to set up the temp dir in more cases...

dogversioning (approved):
Just for my own edification - are there any kind of race condition/overwriting concerns this adds?

mikix:
> Just for my own edification - are there any kind of race condition/overwriting concerns this adds?

There *could* be if we ever go multiprocess. But right now we only have a light smattering of asyncio code, which wouldn't race each other on this.

But also even if we switch to multiprocess, because we added the "atomic writes" bits to avoid half-written files to S3, that would also save us here - at worse two processes would download the same file and one of them would win the atomic write fight-off. But no data would be partially/incorrectly saved to disk.

PR #259, opened 2023-08-07
fix: use separate reading/writing progress bars during tasks

mikix:
This allows for tasks whose output is not 1:1 with their input. For example, medicationrequest or covid_symptom__nlp_results.

Before, their total batch count might be over or under counted. Now, the reading bar will always match the input progress, and the writing bar will be indeterminate or finished while taking a reading breather.

# Considerations
I could have just left off the `Writing` bar, and had the `Reading` bar have momentary-but-unexplained pauses. But then you have the issue of the last batch written. I didn't want 100% on the reading bar for 30 minutes with no explanation (or arguably worse, save a bit at the end and leave it at 99% for 30 minutes).

I also figured it was better to reset the existing `Writing` bar (rather than have a growing list of `Batch 1`, `Batch 2` etc bars) because that kept the Reading bar as the focus and seeing historical batch writes wouldn't be super interesting to the user, I figured. (though sometimes interesting to me for their timings -- now if I want to watch the timing, I have to notice it during a pause in the action... so some possible improvements there).

But I'm open to those thoughts ^ being wrong.

# Screencast of new bars in action:
[Screencast from 08-07-2023 02:53:19 PM.webm](https://github.com/smart-on-fhir/cumulus-etl/assets/1196901/7be0fcca-4a2b-49fb-a481-4d8917353c64)

### Checklist

### DIFF
--- a/cumulus_etl/etl/studies/covid_symptom/covid_tasks.py
+++ b/cumulus_etl/etl/studies/covid_symptom/covid_tasks.py
@@ -6,6 +6,7 @@ import os
 
 import ctakesclient
 import pyarrow
+import rich.progress
 
 from cumulus_etl import common, formats, nlp, store
 from cumulus_etl.etl import tasks
@@ -82,14 +83,14 @@ class CovidSymptomNlpResultsTask(tasks.EtlTask):
         with common.NdjsonWriter(error_path, "a") as writer:
             writer.write(docref)
 
-    async def read_entries(self) -> tasks.EntryIterator:
+    async def read_entries(self, *, progress: rich.progress.Progress = None) -> tasks.EntryIterator:
         """Passes clinical notes through NLP and returns any symptoms found"""
         phi_root = store.Root(self.task_config.dir_phi, create=True)
 
         # one client for both NLP services for now -- no parallel requests yet, so no need to be fancy
         http_client = nlp.ctakes_httpx_client()
 
-        for docref in self.read_ndjson():
+        for docref in self.read_ndjson(progress=progress):
             # Only bother running NLP on docs that are current & finalized
             bad_ref_status = docref.get("status") in ("superseded", "entered-in-error")  # status of DocRef itself
             bad_doc_status = docref.get("docStatus") in ("preliminary", "entered-in-error")  # status of clinical note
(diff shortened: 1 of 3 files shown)


dogversioning:
> I also figured it was better to reset the existing Writing bar (rather than have a growing list of Batch 1, Batch 2 etc bars) because that kept the Reading bar as the focus and seeing historical batch writes wouldn't be super interesting to the user, I figured. 

How would you feel about changing the Writing label as it goes? 'Writing batch 1' in progress, 'Wrote batch 1' when completed, and then 'all batches written' when done?

cumulus_etl/etl/studies/covid_symptom/covid_tasks.py
```diff
@@ -82,14 +83,14 @@ def add_error(self, docref: dict) -> None:
         with common.NdjsonWriter(error_path, "a") as writer:
             writer.write(docref)
 
-    async def read_entries(self) -> tasks.EntryIterator:
+    async def read_entries(self, *, progress: rich.progress.Progress = None) -> tasks.EntryIterator:
```
dogversioning, line 86: is this * just futureproofing?

dogversioning (approved):

cumulus_etl/etl/studies/covid_symptom/covid_tasks.py
mikix: Partly, but also to enforce current kwarg practices. For things like this that don't feel like the "noun" object of the "verb" of the method -- i.e. not `write_entries(rows)` but rather `write_entries(progress)` -- I prefer to forcibly relegate them to the kwargs portion with `*`. I think it helps readability, for myself anyway.

dogversioning: gotcha - not my personal pref but a perfectly defensible thing

mikix:
> How would you feel about changing the Writing label as it goes? 'Writing batch 1' in progress, 'Wrote batch 1' when completed, and then 'all batches written' when done?

We talked on slack about this - I changed the code to say `Writing batch 1`, but not the past tense versions, as that didn't fit the other existing progress bars which don't change tense. It looks good without the change anyway - the 100% indicates we are done with it and the changing number does give clarity for why the bar resets.

PR #260, opened 2023-08-08
feat!: correctly delete old symptoms if NLP now returns zero

mikix:
The goal of this commit is to handle the following case:
- Yesterday you ran NLP and got five symptoms.
- Today you run NLP and get zero symptoms.

Before this commit, the previous five symptoms would remain in Athena. With this commit, they are all deleted from Athena.

This may not sound super common, but it does happen if you change the NLP pipeline (like during the development of an NLP task) or the note text changes.

The ETL currently handles *reduction* of symptoms by defining a group field of `docref_id` and any symptoms that match that group but aren't in the current batch get deleted. But this previously relied on at least one row being present for any group to get noticed.

So if you have zero rows, your group would be left alone, even though NLP says all the group members should be deleted.

Changes in this commit:
- Tasks that use this group feature now must also keep track of the group values as it generates rows.
- Because the `convert` command would not normally see these group definitions (because there is no output ndjson row for them), we now write out a metadata file for each ndjson batch. So for a file like `table.000.ndjson`, there can now be a sibling file `table.000.meta` which holds batch metadata, currently just this groups definition. The `convert` command now passes that along to deltalake.
- Rather than also fix the parquet output format to write out this metadata, I just removed the format. I didn't _need_ to do this, but I figured we don't need this format anymore. If you don't care about human readability, you should just be using the deltalake format. This format has outlived its purpose, so here's a bit of cleanup.
- Now the only remaining output formats are ndjson and deltalake.
- Because I added this breaking change (which should affect no one), I finally bumped our version to 1.0.0 -- this reflects not just this change but the general production-readiness of Cumulus ETL. We still aren't doing actual releases, so semantic versioning is not 100% useful here, but it's just a possible indication to close-watchers that something big has changed.

### Checklist


cumulus_etl/etl/studies/covid_symptom/covid_tasks.py
```diff
@@ -118,11 +122,25 @@ async def read_entries(self, *, progress: rich.progress.Progress = None) -> task
                 self.add_error(orig_docref)
                 continue
 
+            # Record this docref as processed (so we can delete it from our table if we don't generate symptoms).
+            #
+            # We could move this up to the beginning of the loop and delete symptoms for notes now marked as
+            # superseded or are no longer considered an ED docref, for database cleanliness.
+            # But the current approach instead focuses purely on accuracy and makes sure that we zero-out any dangling
+            # entries for groups that we do process.
+            # Downstream SQL can ignore the above cases itself, as needed.
```
dogversioning, line 131: ⚠️⚠️  - do you have a run someplace where all the covid NLP data is zeroed out? can we make sure the library still builds successfully?

dogversioning (approved):

cumulus_etl/etl/studies/covid_symptom/covid_tasks.py
mikix: 1) No I don't right now
2) I wouldn't worry about the entire database going to zero - that would be a true oddity. This PR is more worried about a single document's symptoms going to zero.
3) My comments about downstream SQL there are more like "if document A suddenly gets marked as entered-in-error, let's not bother deleting their symptoms in Athena -- the study should be ignoring entered-in-error docs anyway"

I would expect no changes in the likelihood of the library building or not - this is more like, a couple symptoms might disappear from the database.

dogversioning: got it - makes sense

PR #262, opened 2023-08-09
feat: add new llama2 NLP test task

mikix:
This task is not run by default, and it is more of a toy than a production-ready task right now.

It just asks Llama2 to summarize clinical notes.
There was no prompt engineering to improve results, yet.

Here's how you run it:
```
export HUGGING_FACE_HUB_TOKEN=xxx
docker compose up llama2 --wait  # this will take several minutes
docker compose run cumulus-etl --task hftest__summary ...
```

Note that Llama2 requires a GPU to run.

### Checklist

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -70,6 +70,31 @@ services:
           devices:
             - capabilities: [gpu]
 
+  llama2:
+    image: ghcr.io/huggingface/text-generation-inference:1.0.0
+    environment:
+      # If you update anything here that could affect NLP results, consider updating the
+      # task_version of any tasks that use this docker.
+      - HUGGING_FACE_HUB_TOKEN
+      - MAX_BATCH_PREFILL_TOKENS=2048  # default of 4096 overwhelms a 16GB machine
+      - MODEL_ID=meta-llama/Llama-2-7b-chat-hf
+      - REVISION=08751db2aca9bf2f7f80d2e516117a53d7450235
+    healthcheck:
+      # There's no curl or wget inside this container, but there is python3!
+      test: ["CMD", "python3", "-c", "import socket; socket.create_connection(('localhost', 80))"]
+      start_period: 20m  # give plenty of time for startup, since we may be downloading a model
+    volumes:
+      - hf-data:/data
+    networks:
+      - cumulus-etl
+    ports:
+      - 8086:80
+    deploy:
+      resources:
+        reservations:
+          devices:
+            - capabilities: [gpu]
+
   # these images are intended specifically for development work
   cumulus-etl-test:
     extends:
@@ -156,3 +181,4 @@ networks:
 
 volumes:
   ctakes-overrides:
+  hf-data:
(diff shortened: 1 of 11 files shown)


compose.yaml
```diff
@@ -70,6 +70,31 @@ services:
           devices:
             - capabilities: [gpu]
 
+  llama2:
```
dogversioning, line 73: should there be a new profile for when you want to spin up the ETL with this?

cumulus_etl/etl/studies/hftest/hf_tasks.py
dogversioning, line 1: this is not really a note on this particular task and is more just pontificating for the future, but it would be neat if the final llama task could be kind of generic, as opposed to the covid one which is very specific.

dogversioning (approved):

compose.yaml
mikix: I wanted to avoid profiles and keep this manual for now, until usage is more obvious...

I agree it probably shouldn't go in the normal support profile yet/ever because the resource profile on this will be so intense. You likely have a custom machine for llama2, unlike normal ETL where you can get away with doing everything locally.

But separately... I'm starting to feel the inflexibility of offloading the responsibility of this docker compose stuff to the user. Like, the other services (ctakes & cnlp) are just in service of one task (so far), and our instructions have you start them regardless of whether you need them at all. As we add more like that, it would be nice if the ETL sat above/next-to the docker compose layer and could also drive it, starting and stopping services.  (that doesn't really jive with the "llama2 on a separate machine" idea above, but just a thought)

dogversioning: it might be worth looking into docker swarm? not as part of this, of course, but allows for some dynamism in container spinup.

cumulus_etl/etl/studies/hftest/hf_tasks.py
dogversioning: I guess i'm thinking 'base llama is generic, study prompts are specific' - but that may be wishful thinking. Regardless, nothing needs to happen on this thread now.

mikix: Ah maybe that's just a confusion about how I named this `hftest` task? I think our brains are aligned in this case.

Task == study prompt. You could imagine a future with five or so tasks, each with their own prompt. "base llama" is the code I added in `nlp/huggingface.py` which handles the logic of talking to a llama. But any task can call that with its own prompt.

dogversioning: ok yeah we're at tanagra.

mikix: I do think there's room to further reduce NLP task boilerplate, to let each study task really really just focus on the prompt (pull out the looping over docrefs and standard filtering). But I think I'll bite that off a bit later, as overlap becomes more obvious.

The idea above about a mode/product that lets you play with prompting over some sample clinical notes does feel like a good thing to build at some point. It's not crazy to put that into the `chart-review` portion of the ETL maybe? I dunno - we need to further explore that use case.

PR #265, opened 2023-08-15
feat: show time elapsed rather than attempting to estimate remaining

mikix:
### Checklist

### DIFF
--- a/cumulus_etl/etl/tasks/base.py
+++ b/cumulus_etl/etl/tasks/base.py
@@ -153,7 +153,7 @@ class EtlTask:
         """Writes all entries to each output tables in batches"""
 
         def update_status():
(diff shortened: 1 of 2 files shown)


cumulus_etl/etl/tasks/base.py
```diff
@@ -153,7 +153,7 @@ async def _write_tables_in_batches(
         """Writes all entries to each output tables in batches"""
 
         def update_status():
-            status.plain = "\n".join(f"{x.success:,} processed for {x.label}" for x in self.summaries)
+            status.plain = "\n".join(f"{x.success:,} written to {x.label}" for x in self.summaries)
```
mikix, line 156: Kind of unrelated, but I felt like this messaging is more accurate, especially as you see the "reading" bar climb up and not the "processed" number

PR #271, opened 2023-08-29
feat: give more memory to the Delta Like workers

mikix:
We were noticing that after about 60M rows, a delta lake table can no longer be merged to, with only 2G of driver memory.

So bumping it to 4G for extra room. This may mean that batch sizes should be adjusted downward, as the driver takes more system memory.

I will notify folks in our slack about this possible adjustment.

### Checklist

### DIFF
--- a/docs/performance.md
+++ b/docs/performance.md
@@ -32,14 +32,20 @@ So the larger it is, the more memory will be used and the larger the output file
 
 ### Recommended Setup
 
-Use 16GB of memory with `--batch-size=300000`.
-We've never found it to run out of memory that way, and it should still have some leeway for spikes of memory use.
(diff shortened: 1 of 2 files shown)


docs/performance.md
```diff
@@ -32,14 +32,20 @@ So the larger it is, the more memory will be used and the larger the output file
 
 ### Recommended Setup
 
-Use 16GB of memory with `--batch-size=300000`.
-We've never found it to run out of memory that way, and it should still have some leeway for spikes of memory use.
+We've found `--batch-size=100000` works well for 16GB of memory.
+And `--batch-size=500000` works well for 32GB of memory.
```
mikix, line 36: These are what I've been using. :shrug:

dogversioning (approved):

PR #274, opened 2023-09-22
feat(chart-review): group notes by encounter, show real & anon IDs

mikix:
This adds a couple quality-of-life improvements to chart-review mode:
- Notes are now grouped by encounter, with each note clearly demarked inside a larger text document.
- Notes now have a header with their note type label (e.g. "Admission MD") plus their real & anonymized IDs
- The encounter ID (real+anon) plus all the contained DocRef IDs (real+anon) are also pushed as Label Studio metadata, which will show up in exports if folks want to do some post-processing.

This should land along with an announcement.

### Checklist

### DIFF
--- a/cumulus_etl/chart_review/cli.py
+++ b/cumulus_etl/chart_review/cli.py
@@ -29,7 +29,7 @@ def init_checks(args: argparse.Namespace):
 
 
 async def gather_docrefs(
-    client: fhir.FhirClient, root_input: store.Root, root_phi: store.Root, args: argparse.Namespace
+    client: fhir.FhirClient, root_input: store.Root, codebook: deid.Codebook, args: argparse.Namespace
 ) -> common.Directory:
     """Selects and downloads just the docrefs we need to an export folder."""
     common.print_header("Gathering documents...")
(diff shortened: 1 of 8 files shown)


tests/fhir/test_fhir_client.py
```diff
@@ -227,7 +227,7 @@ async def test_bad_smart_config(self, bad_config_override):
             {},
             (
                 "Client error '400 Bad Request' for url 'https://auth.example.com/token'\n"
-                "For more information check: https://httpstatuses.com/400"
+                "For more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400"
```
mikix, line 230: Unrelated, but httpx changed its output format. Maybe we should be more resilient to that, but part of me likes the strictness and doesn't mind updating once in a blue moon.

cumulus_etl/chart_review/cli.py
```diff
@@ -94,6 +115,49 @@ def philter_notes(notes: Collection[LabelStudioNote], args: argparse.Namespace)
...
 
 
+def group_notes_by_encounter(notes: Collection[LabelStudioNote]) -> list[LabelStudioNote]:
+    """
+    Gather all notes with the same encounter ID together into one note.
+
+    We also add little headers to each note with the Real & Anonymized docref IDs, to help reviewers.
+    We've found that reviewers appreciate both practices.
+    """
+    big_notes = []
```
dogversioning, line 125: suggestion, feel free to disregard (especially because, i, personally, find **BIG NOTES** hilarious for reasons): concatenated_notes might be more descriptive? this trickles down through the rest of this function.

tests/fhir/test_fhir_client.py
dogversioning: honestly i don't hate it either, and that's probably a better source for that info anyway.

dogversioning (approved):

cumulus_etl/chart_review/cli.py
mikix: :smile: I changed the `big_` prefix in this method to `grouped_` which felt easier to say and matches the function name too (`group_notes_by_encounter`).

PR #275, opened 2023-09-26
feat: add new covid_symptom__nlp_results_term_exists task

mikix:
- Adds a new covid_symptom__nlp_results_term_exists task, which uses
  the "termexists" model for polarity checking cTAKES rather than the
  previous "negation" model. This task will largely be used to compare
  the performance of the two models.
- Rename some docker compose targets, like the etl-support profile
  into the covid-symptom profile (the thinking is that we'll have
  study-specific sets of services that you might want on or off,
  depending on what you're doing)
- Refactors some NLP code to use shared base classes.
- Remove covid_symptom__nlp_results from the default set of tasks.
  Study-specific tasks should have to be requested.

### Checklist


compose.yaml
```diff
@@ -49,20 +51,39 @@ services:
   ctakes-covid:
     extends: ctakes-covid-base
     profiles:
-      - etl-support
-      - etl-support-gpu
+      - covid-symptom
+      - covid-symptom-gpu
```
mikix, line 55: OK notable change number 1: how do you feel about this direction?

I think I mentioned wanting to toy with this - not a global "run all the services I might ever need" but rather a more targeted "run the services for the studies I'm caring about right now".

It's a more complicated user story, but I feel like the "run everything" approach quickly falls over as we add studies.

You had mentioned anticipating moving to a more comprehensive meta-runner like Kubernetes or something yeah? I'm not opposed, but in the meantime, I'm thinking something this could be a happy medium.

cumulus_etl/etl/studies/covid_symptom/covid_tasks.py
```diff
@@ -58,42 +57,13 @@ def is_ed_docref(docref):
     return any(is_ed_coding(x) for x in codings)
 
 
-class CovidSymptomNlpResultsTask(tasks.EtlTask):
-    """Covid Symptom study task, to generate symptom lists from ED notes using NLP"""
+class BaseCovidSymptomNlpResultsTask(tasks.BaseNlpTask):
+    """Covid Symptom study task, to generate symptom lists from ED notes using cTAKES + a polarity check"""
```
mikix, line 61: A lot of this PR is refactoring, sorry.

I make a new BaseNlpTask for all NLP jobs (shared by the hugging face test task and the covid tasks) PLUS a base covid symptoms test here that both the (nearly-identical) covid tasks inherit from).

And the tests when you get to them are also refactored a bit - I added a new shared BaseEtlTest class for all the tests that care about running a full ETL and some refactoring to handle the changes from dropping the covid tests from the default list of ETL jobs (keep reading for that change).

cumulus_etl/etl/tasks/nlp_task.py
```diff
@@ -0,0 +1,83 @@
...
+import os
+from typing import Callable
+
+import rich.progress
+
+from cumulus_etl import common, fhir, nlp, store
+from cumulus_etl.etl.tasks.base import EtlTask, OutputTable
+
+
+class BaseNlpTask(EtlTask):
```
mikix, line 14: This is the base NLP task I mentioned before. It basically just handles looping over the DocRefs and extracting the notes. With some boilerplate code for handling group fields if the subclass wants that.

docs/studies/covid-symptom.md
mikix, line 1: I figured if the docker compose has study specific commands now, we might want a per-study doc about how to run the study or other study-specific notes.

cumulus_etl/etl/cli.py
```diff
@@ -68,15 +68,15 @@ def check_mstool() -> None:
...
     """
     Verifies that all external services and programs are ready
 
     May block while waiting a bit for them.
     """
-    nlp.check_ctakes()
-    nlp.check_cnlpt()
     check_mstool()
+    for task in tasks:
+        await task.init_check()
```
mikix, line 79: Because I the covid support services like ctakes/cnlpt are now study-specific, I figured it made sense to finally add the ability for the selected tasks to check for the services they'll need.

So now the covid tasks check for ctakes and the cnlpt models it needs.

docs/studies/covid-symptom.md
```diff
@@ -0,0 +1,51 @@
+---
+title: Covid Symptom
+parent: Studies
+grand_parent: ETL
+nav_order: 1
+# audience: engineer familiar with the project
+# type: howto
+---
+
+# The Covid Symptom Study
```
mikix, line 10: I figured if the docker compose has study specific commands now, we might want a per-study doc about how to run the study or other study-specific notes.

cumulus_etl/etl/cli.py
```diff
@@ -68,15 +68,15 @@ def check_mstool() -> None:
...
     """
     Verifies that all external services and programs are ready
 
     May block while waiting a bit for them.
     """
-    nlp.check_ctakes()
-    nlp.check_cnlpt()
     check_mstool()
+    for task in selected_tasks:
+        await task.init_check()
```
mikix, line 79: Because I the covid support services like ctakes/cnlpt are now study-specific, I figured it made sense to finally add the ability for the selected tasks to check for the services they'll need.

So now the covid tasks check for ctakes and the cnlpt models it needs.

compose.yaml
dogversioning: yeah i don't hate this given the addition of a new task, and also that folks may not often be running this now.

cumulus_etl/etl/tasks/nlp_task.py
```diff
@@ -0,0 +1,83 @@
...
+    # The "task_version" field is a simple integer that gets incremented any time an NLP-relevant parameter is changed.
+    # This is a reference to a bundle of metadata (model revision, container revision, prompt string).
+    # We could combine all that info into a field we save with the results. But it's more human-friendly to have a
+    # simple version to refer to.
+    #
+    # CONSIDERATIONS WHEN CHANGING THIS:
+    # - Record the new bundle of metadata in your class documentation
+    # - Update any safety checks in prepare_task() or elsewhere that check the NLP versioning
+    # - Be aware that your caching will be reset
+    task_version = 1
```
dogversioning, line 34: if you want this to start at 0 (like in hf_tasks), this should also start as 0?

docs/studies/covid-symptom.md
```diff
@@ -0,0 +1,51 @@
...
+
+This will need access to clinical notes,
+which are pulled fresh from your EHR (since the ETL doesn't store clinical notes).
+This means you will likely have to provide some other FHIR authentication arguments like
+`--smart-client-id` and `--fhir-url`.
+See `--help` for more authentication options.
+
+There is also a second optional task `--task covid_symptom__nlp_results_term_exists`,
+which just uses a different polarity cNLP transformer (`termexists` rather than `negation`).
+You likely don't need both, but they may be interesting to compare.
```
dogversioning, line 51: Q: do we want to treat this as a in development style command? i.e. maybe it doesn't live here until we've decided to use it in a non-experimental way?

dogversioning (approved):

cumulus_etl/etl/tasks/nlp_task.py
mikix: I was mentally treating 0 vs 1 like semantic versioning does - that at 1 you are opting into actually caring about breaks but at zero you don't need to care about them.

So I was just thinking I'd want to push subclasses to be well-versioned from the start (hf_test is a toy thing that will never be a real boy, so it doesn't matter).

docs/studies/covid-symptom.md
mikix: Good question.

So this task is not like, in-development or experimental in the sense of a WIP that's likely to change. Once this lands, I feel like it's a full, normal task. But it is "less than" a normal task because it's kind of a special case task to help Tim compare the new transformer's performance. But I guess all study tasks are kind of special case tasks like that.

But here's this task's current status in this PR:
1. is not run by default
2. can only be run by name (i.e. not tagged at all -- `--task-filter=covid_symptom` or `--task-filter=gpu` do not hit it)
3. is documented on the study's docs

Is your thinking that 1 and 2 are good (i.e. "lowlightling" this task) and 3 should also not mention the task to lowlight it even further?

mikix: (I'm open to that, just trying to clarify what you are thinking about when marking experimental -- or do you want like active barriers like a flag `--enable-experimental` or something)

dogversioning: i think in my head, i'm drawing the line at 'this is something we are using internally while doing study validation' versus 'this is something a partner site might be explicitly asked to run' - and I think this is mostly a 3 level comment.

mikix: Fair yeah - I dropped mention from the docs :+1:

PR #279, opened 2023-10-09
docs: emphasize LLMs and re-usable code in NLP docs

mikix:
### Checklist

### DIFF
--- a/docs/nlp.md
+++ b/docs/nlp.md
@@ -17,13 +17,14 @@ and records those findings alongside the coded FHIR data.
 
 This way you can often surface symptoms that simply aren't recorded in the traditional FHIR data.
 
-## NLP Is Always Specific to a Study
+## NLP Is Always Specific to a Clinical Purpose
 
 The first thing to understand is that Cumulus ETL always
-runs NLP in the context of a specific study purpose.
+runs NLP in the context of a specific clinical purpose,
+which we'll call a "study."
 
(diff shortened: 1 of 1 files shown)


docs/nlp.md
```diff
@@ -17,22 +17,109 @@ and records those findings alongside the coded FHIR data.
 
 This way you can often surface symptoms that simply aren't recorded in the traditional FHIR data.
 
-## NLP Is Always Specific to a Study
+## NLP Is Always Specific to a Clinical Purpose
```
mikix, line 20: Andy bumped on this, so I'm softening this language a bit

docs/nlp.md
```diff
@@ -17,22 +17,109 @@ and records those findings alongside the coded FHIR data.
...
+
+Those calls to `etl.*` are calls to the internal NLP support code that the task does not have to
+re-invent.
+
+And with that relatively low level of complexity (though finding the right prompt can be hard),
+you've got a study task that you can run over all your institution's clinical notes. 
+
 ## Available NLP Strategies
 
+### Large Language Models (LLMs)
```
mikix, line 70: Moved this section above cTAKES and talk more about local & cloud LLMs in general, to emphasize our support here.

Unfortunately, we don't actually have shipped code for a cloud LLM or this section could be stronger. But we can update it when we do.

dogversioning (approved):

PR #283, opened 2023-10-19
feat: adds four new default FHIR resources to the mix

mikix:
New & default tasks:
- allergyintollerance
- device
- diagnosticreport
- immunization

I've also modified some of the existing unit testing to be more specific about which tests to run, in the cases where it doesn't make sense to update tests when you add a new task.

### Checklist


cumulus_etl/deid/ms-config.json
```diff
@@ -47,7 +47,6 @@
     {"path": "nodesByType('ContactDetail')", "method": "redact"},
     {"path": "nodesByType('ContactPoint')", "method": "redact"},
     {"path": "nodesByType('HumanName')", "method": "redact"},
-    {"path": "nodesByType('Identifier')", "method": "redact"},
```
mikix, line 50: This was previously a blanket injunction against using `Identifier` elements, because I'd only seen them in `Resource.identifier` contexts, where the EHR stores PHI identifiers like MRNs or other such business-case oriented identifiers (I think in Patients, the SSN goes there? Or maybe that's an extension, I don't remember.)

Anyway. Device has a version.component field that for some reason uses the `Identifier` element, so I can't blanket-ban it now. But it is also never allow-listed anywhere else. This line was just a second line of defense against mistakes. But now that we have a real use case, we can drop this.

cumulus_etl/deid/ms-config.json
```diff
@@ -156,6 +228,43 @@
...
+    {"path": "Immunization.site", "method": "keep"},
+    {"path": "Immunization.route", "method": "keep"},
+    {"path": "Immunization.doseQuantity", "method": "keep"},
+    {"path": "Immunization.performer.function", "method": "keep"},
+    {"path": "Immunization.performer.actor", "method": "keep"},
+    {"path": "Immunization.reasonCode", "method": "keep"},
+    {"path": "Immunization.reasonReference", "method": "keep"},
+    {"path": "Immunization.isSubpotent", "method": "keep"},
+    {"path": "Immunization.subpotentReason", "method": "keep"},
+    {"path": "Immunization.education.publicationDate", "method": "keep", "comment": "we aren't keeping documentType or uri for fear of any PHI slipping in, so maybe these two dates aren't very useful, but better more data than less"},
```
mikix, line 255: I'm still not 100% on whether it's safe to include `documentType`/`uri` -- they seem kinda harmless, but I also don't think they would be very valuable... :shrug:

tests/data/mstool/output/AllergyIntolerance.ndjson
```diff
@@ -0,0 +1,30 @@
+{
+  "resourceType" : "AllergyIntolerance",
+  "meta": {
+    "security": [{
+      "code": "REDACTED",
+      "display": "redacted",
+      "system": "http://terminology.hl7.org/CodeSystem/v3-ObservationValue"
+    }]
+  },
```
dogversioning, line 9: mostly for my own edification: how does this work on the MS side?

tests/data/mstool/output/AllergyIntolerance.ndjson
```diff
@@ -0,0 +1,30 @@
...
+  "recordedDate" : "2023",
+  "recorder" : { "reference": "Practitioner/x" },
+  "asserter" : { "reference": "Patient/x" },
+  "lastOccurrence" : "2023",
+  "reaction" : [{
+    "substance" : { "text": "kept" },
+    "manifestation" : [{ "text": "kept" }],
+    "onset" : "2023",
+    "severity" : "mild",
+    "exposureRoute" : { "text": "kept" }
```
dogversioning, line 28: also `note`?

dogversioning (approved):

tests/data/mstool/output/AllergyIntolerance.ndjson
mikix: Ah OK for context: this file is the expected output of the MS tool. It's input counterpart is in the `input/` folder.

To your actual question: this bit here is inserted by the MS tool automatically (can't turn it off). I find this annoying, and pointless for Cumulus' purposes, so the ETL drops this field again when processing files.

But this unit test doesn't do the full ETL processing, it's just testing our ms-config.json file, so the field is still present.

tests/data/mstool/output/AllergyIntolerance.ndjson
mikix: `note` is dropped per the `ms-config.json` config. So it doesn't show up in this "expected output" file. Whether we should drop it or not, is open for discussion. But we tend to drop freeform text fields like that because of high risk of PHI in them, and it's not machine-readable, unless we wanted to NLP it.

tests/data/mstool/input/DiagnosticReport.ndjson
```diff
@@ -0,0 +1,24 @@
+{
+  "resourceType" : "DiagnosticReport",
+  "identifier" : [{ "value": "dropped" }],
+  "basedOn" : [{ "reference": "CarePlan/x" }],
+  "status" : "final",
+  "category" : [{ "text": "kept" }],
+  "code" : { "text": "kept" },
+  "subject" : { "reference": "Patient/x" },
+  "encounter" : { "reference": "Encounter/x" },
+  "effectiveDateTime" : "2023",
```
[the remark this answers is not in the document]
mikix: For "variant" fields, I don't usually unit-test each possible variation. Only the "interesting" variations (like if we drop the string variant, but keep the others, I'll have a unit test for an arbitrary one that we keep and also a unit test for the dropped variant).

We can't put multiple variants in one test file because the MS tool would complain about invalid FHIR.

cumulus_etl/deid/ms-config.json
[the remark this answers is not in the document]
mikix: Yeah this is the juicier file for sure :smile: were all the redaction decisions live.

You mentioned `note` and `effectivePeriod`:
- `AllergyIntolerance.note` is missing from this file (not allow-listed) intentionally - but that's open to discussion.
- `DiagnosticReport.effectivePeriod` is picked up by keeping `DiagnosticReport.effective` which keeps all the variants of that field.

Happy to talk more about each line in here - honestly, for safety's sake, maybe it's worth you and I going over this file's changes line by line sometime next week.

PR #285, opened 2023-10-26
feat(chart-review): push internal docref spans to Label Studio

mikix:
In case we need to know the internal placing of doc notes, when we gather multiple notes into one note, push up the start & end of those texts.

No planned use for this right now, but I want to get this in so when we do want it, the metadata already exists.

### Checklist

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -51,12 +51,15 @@ services:
   ctakes-covid:
     extends: ctakes-covid-base
     profiles:
+      - chart-review
+      - chart-review-gpu
       - covid-symptom
       - covid-symptom-gpu
 
   cnlpt-negation:
     image: smartonfhir/cnlp-transformers:negation-0.6.1-cpu
     profiles:
+      - chart-review
       - covid-symptom
     networks:
       - cumulus-etl
@@ -64,6 +67,7 @@ services:
   cnlpt-negation-gpu:
     image: smartonfhir/cnlp-transformers:negation-0.6.1-gpu
     profiles:
+      - chart-review-gpu
       - covid-symptom-gpu
     networks:
       - cumulus-etl
--- a/cumulus_etl/chart_review/cli.py
+++ b/cumulus_etl/chart_review/cli.py
@@ -20,7 +20,7 @@ def init_checks(args: argparse.Namespace):
 
     if args.nlp:
         nlp.check_ctakes()
-        nlp.check_cnlpt()
+        nlp.check_negation_cnlpt()
 
     if not cli_utils.is_url_available(args.label_studio_url, retry=False):
         errors.fatal(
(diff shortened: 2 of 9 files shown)


cumulus_etl/chart_review/cli.py
```diff
@@ -101,7 +111,12 @@ async def run_nlp(notes: Collection[LabelStudioNote], args: argparse.Namespace)
         ctakes_json = await ctakesclient.client.extract(note.text, client=http_client)
         matches = ctakes_json.list_match(polarity=Polarity.pos)
         spans = ctakes_json.list_spans(matches)
-        cnlpt_results = await ctakesclient.transformer.list_polarity(note.text, spans, client=http_client)
+        cnlpt_results = await ctakesclient.transformer.list_polarity(
+            note.text,
+            spans,
+            client=http_client,
+            model=nlp.TransformerModel.NEGATION,
```
mikix, line 118: I just wanted to be more explicit about which transformer we're using in chart-review mode, so we prevent surprise changes to the default model.

cumulus_etl/chart_review/downloader.py
```diff
@@ -68,7 +68,7 @@ async def _download_docrefs_from_real_ids(
 
     # Grab identifiers for which specific docrefs we need
     with common.read_csv(docref_csv) as reader:
-        docref_ids = {row["docref_id"] for row in reader}
+        docref_ids = sorted({row["docref_id"] for row in reader})
```
mikix, line 71: To aid testing / reproducability

cumulus_etl/chart_review/labelstudio.py
```diff
@@ -17,14 +18,24 @@
...
+
+    # A title is only used when combining notes into one big encounter note. It's not sent to Label Studio.
+    title: str = ""
+
+    # Doc mappings is a dict of real DocRef ID -> anonymized DocRef ID of all contained notes, in order
+    doc_mappings: dict[str, str] = dataclasses.field(default_factory=dict)
+
+    # Doc spans indicate which bits of the text come from which DocRef - it will map real DocRef ID to a pair of
+    # "first character" (0-based) and "last character" (0-based, exclusive) - just like cTAKES match text spans.
+    doc_spans: dict[str, tuple[int, int]] = dataclasses.field(default_factory=dict)
```
mikix, line 35: This is the new bit - rest is just documenting / defaulting

compose.yaml
```diff
@@ -51,19 +51,23 @@ services:
   ctakes-covid:
     extends: ctakes-covid-base
     profiles:
+      - chart-review
+      - chart-review-gpu
```
mikix, line 55: I realized that with the previous switch to study-specific profiles, the `chart-review` mode, which has its own needs (ctakes and the negation transformer) didn't have a profile, but that it should - like a study does.

Doc changes were also made lower down

cumulus_etl/chart_review/labelstudio.py
dogversioning: this is an aside, but i do like the dataclass format for readability.

dogversioning (approved):

PR #289, opened 2023-11-10
fix: write "null match" NLP rows for DocRefs with no symptoms

mikix:
Previously, there was no way to detect the difference between a DocRef that was not processed (not an ED DocRef or an error occurred etc) vs a DocRef that had no symptoms.

Both just had zero rows in the Athena db.

This commit starts adding a single row in the nlp_results table with a null "match" column for this case.

Existing SQL should be able to gracefully handle this change. If/when we want to update the SQL to handle these kinds of DocRefs, it can be updated to treat these new rows specially. But for now, they will be ignored.

### Checklist

### DIFF
--- a/cumulus_etl/etl/studies/covid_symptom/covid_ctakes.py
+++ b/cumulus_etl/etl/studies/covid_symptom/covid_ctakes.py
@@ -86,20 +86,28 @@ async def covid_symptoms_extract(
         logging.warning("Could not check polarity for docref %s (%s): %s", docref_id, type(exc).__name__, exc)
         return None
 
+    # Helper to make a single row (match_value is None if there were no found symptoms at all)
+    def _make_covid_symptom_row(row_id: str, match: dict | None) -> dict:
+        return {
+            "id": row_id,
+            "docref_id": docref_id,
+            "encounter_id": encounter_id,
+            "subject_id": subject_id,
+            "generated_on": timestamp,
+            "task_version": task_version,
+            "match": match,
+        }
+
     # Now filter out any non-positive matches
     positive_matches = []
     for i, match in enumerate(matches):
         if polarities_cnlp[i] == ctakesclient.typesystem.Polarity.pos:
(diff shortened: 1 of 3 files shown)


docs/index.md
```diff
@@ -21,4 +21,9 @@ ETL stands for "extract, transform, load."
...
 
 Read the [First Time Setup](setup) documentation to learn how to install & prepare Cumulus ETL
-at your institution.
\ No newline at end of file
+at your institution.
+
+## Source Code
+Cumulus ETL is open source.
+If you'd like to browse its code or contribute changes yourself,
+the code is on [GitHub](https://github.com/smart-on-fhir/cumulus-etl).
```
mikix, line 29: Unrelated, but wanted to squeeze this in after we talked about it

cumulus_etl/etl/studies/covid_symptom/covid_ctakes.py
```diff
@@ -86,20 +86,28 @@ def is_covid_match(m: ctakesclient.typesystem.MatchText):
...
-                    "match": match.as_json(),
-                }
-            )
+            positive_matches.append(_make_covid_symptom_row(f"{docref_id}.{i}", match.as_json()))
+
+    if not positive_matches:
+        # In this case, we write out a single row with a null match,
+        # to flag to the downstream SQL that this DocRef was processed,
+        # it just had no symptoms (vs a DocRef that we never handled).
+        positive_matches = [_make_covid_symptom_row(f"{docref_id}.0", None)]
```
mikix, line 111: 1. I tested with a Synthea dataset that rows with null matches don't affect our current covid_symptom SQL code at all (they are ignored). So this is harmless and will let us write SQL in future that does care about these docs.
2. The SQL schema for this table is provided manually by the ETL (not inferred), so the SQL will always know the schema, even if for some weird reason all the `match` columns are null.

dogversioning (approved):

PR #290, opened 2023-11-13
fix: expand partial CodeableConcepts in the final FHIR schema

mikix:
If we see any piece of a CodeableConcept (or Coding) in the FHIR input, include the full CodeableConcept/Coding struct type.

In practice, this means code types are either fully present or null.

This is all to make downstream SQL a little easier to write, while still not bothering with a full FHIR schema up and down the element tree.

### Checklist


dogversioning (approved):
thank you so much

PR #292, opened 2024-01-04
fix: hand symptoms to cTAKES in a format it expects

mikix:
Ever since the March 17th, 2023 commit that fed cTAKES a custom bsv file, our COVID NLP performance has taken a hit (unbeknownst to us).

cTAKES actually expects a slightly different format for symptoms bsv files, which we now enforce as we send it files.

This gets our COVID NLP performance on a standard BCH set of 200-odd Cerner html notes from 0.784 to 0.813 F1.

This commit bumps the covid tasks' task_format from 3 to 4.

### Checklist


pyproject.toml
```diff
@@ -18,7 +18,7 @@ dependencies = [
     "inscriptis < 3",
     "jwcrypto < 2",
     "label-studio-sdk < 1",
-    "oracledb < 2",
+    "oracledb < 3",
```
mikix, line 21: Unrelated - just noticed that they released a new major version, dropping support for python 3.6

dogversioning (approved):

PR #294, opened 2024-01-10
feat!: rename chart-review to upload-notes

mikix:
"chart-review" is an overloaded term and conflicts with the name of some tooling we are also crafting for help with post-review statistics.

So let's use upload-notes instead, which is more descriptive of what we are actually doing with the command.

chart-review will stay as an alias for now.

### Checklist


docs/chart-review.md
```diff
@@ -8,7 +8,9 @@ nav_order: 20
 
 # Chart Review
```
mikix, line 9: I kept this documentation file as being *about* Chart Review, but everything inside describes the upload-notes command and what I call "upload mode".

I figured Chart Review was what the user's brain would say they are trying to do, and this doc describes the Cumulus ETL command that helps with that.

But open to notes on that - I could buy confusion around it.

docs/chart-review.md
```diff
@@ -191,36 +193,34 @@ simply pass in a new dictionary like so: `--symptoms-bsv /in/my-symptoms.bsv`.
 
 This file should look like (this is a portion of the default Covid dictionary):
 ```
-##  Columns = CUI|TUI|CODE|SAB|STR|PREF
+##  Columns = CUI|TUI||STR|PREF
```
mikix, line 196: Whoops, should have done this in the recent PR about 4-cols, but doing it here is fine too. We gracefully handle 6-cols anyway, but let's document the intended format.

docs/chart-review.md
dogversioning: i think chart review is the business logic name for this, so i buy it.

dogversioning (approved):
Frankly - I think you could remove the old command now, and it would be the right time, since it's internal only at the moment. Up to you.

mikix:
> Frankly - I think you could remove the old command now, and it would be the right time, since it's internal only at the moment. Up to you.

Fair... But I like _some_ overlap makes me feel good. Maybe drop in 3 months or so.

PR #295, opened 2024-01-10
ci: add full stack NLP regression test

mikix:
This is overdue. We want to make sure that any NLP changes are not suprises.

### Checklist


.github/workflows/ci.yaml
```diff
@@ -57,10 +57,51 @@ jobs:
...
+        uses: docker/setup-buildx-action@v3
+
+      - name: Build ETL image
+        uses: docker/build-push-action@v5
+        with:
+          load: true  # just build, no push
+          tags: smartonfhir/cumulus-etl:latest
+
+      - name: Download NLP images
+        run: docker compose --profile covid-symptom up -d --quiet-pull
```
mikix, line 77: It's not clear how to best cache the pulled images...?  Do you think GitHub does some network-side caching? There are some rando actions in the github action marketplace that promise to cache pulled images... But it might be easy enough to do on our own too.

On the other hand... it's just GitHub's resources we're annoying - which makes me suspect they do some their-side caching, just to save their own money and maybe we don't have to be clever.

.github/workflows/ci.yaml
```diff
@@ -37,7 +37,7 @@ jobs:
           pip install .[tests]
 
       - name: Check out MS tool
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
```
mikix, line 40: I updated a lot of actions here, but nothing interesting - just node 20 upgrades.

cumulus_etl/etl/studies/covid_symptom/covid_ctakes.py
```diff
@@ -73,7 +73,12 @@ async def covid_symptoms_extract(
     def is_covid_match(m: ctakesclient.typesystem.MatchText):
         return bool(covid_symptom_cuis.intersection({attr.cui for attr in m.conceptAttributes}))
 
-    matches = list(filter(is_covid_match, matches))
+    matches = filter(is_covid_match, matches)
+
+    # For better reliability when regression/unit testing, sort matches by begin / first code.
+    # (With stable sorting, we want the primary sort to be done last.)
+    matches = sorted(matches, key=lambda x: x.conceptAttributes and x.conceptAttributes[0].code)
+    matches = sorted(matches, key=lambda x: x.begin)
```
mikix, line 81: This is the only actual code change. Just sorting the ctakes results so we can safely compare over time.

.github/workflows/ci.yaml
dogversioning: I would assume they've got something akin to artifactory instance(s) inside their infrastructure. Since it's open source and we don't pay for time, I don't think we need to worry about optimizing - but if we were, it would probably be something similar - add a cache box and update docker hosts to point at it?

.github/workflows/ci.yaml
```diff
@@ -57,10 +57,51 @@ jobs:
...
+            --output-format=ndjson \
+            --task covid_symptom__nlp_results
+
+          # Compare results
+          export OUTDIR=$DATADIR/run-output/covid_symptom__nlp_results
+          sudo chown -R $(id -u) $OUTDIR
+          sed -i 's/"generated_on": "[^"]*", //g' $OUTDIR/*.ndjson
+          diff -upr $DATADIR/expected-output $OUTDIR
+
+          echo "All Good!"
```
dogversioning, line 99: no emojis? i am SHOCKED

dogversioning (approved):

.github/workflows/ci.yaml
mikix: I thought about it! :smile:

PR #297, opened 2024-02-29
feat(upload-notes): add ability to tag philter spans, not redact them

mikix:
This commit deprecates --no-philter in favor of --philter=disable and adds --philter=label (and the default --philter=redact).

When --philter=label is requested, we send a prediction layer to Label Studio that highlights all the detected PHI spans for you.

This is helpful when doing manual de-identification, since philter can do some of the grunt work for you.

Example (synthea note, no PHI here):
image

### Checklist

### DIFF
--- a/cumulus_etl/upload_notes/cli.py
+++ b/cumulus_etl/upload_notes/cli.py
@@ -12,6 +12,10 @@ from cumulus_etl import cli_utils, common, deid, errors, fhir, nlp, store
 from cumulus_etl.upload_notes import downloader, selector
 from cumulus_etl.upload_notes.labelstudio import LabelStudioClient, LabelStudioNote
 
+PHILTER_DISABLE = "disable"
+PHILTER_REDACT = "redact"
+PHILTER_LABEL = "label"
+
 
 def init_checks(args: argparse.Namespace):
     """Do any external service checks necessary at the start"""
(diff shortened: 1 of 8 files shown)


cumulus_etl/upload_notes/cli.py
```diff
@@ -208,8 +221,16 @@ def define_upload_notes_parser(parser: argparse.ArgumentParser) -> None:
...
+    parser.add_argument(
+        "--philter",
+        choices=[PHILTER_DISABLE, PHILTER_REDACT, PHILTER_LABEL],
+        default=PHILTER_REDACT,
+        help="Whether to use philter to redact/tag PHI",
+    )
+    # Old, simpler version of the above
     parser.add_argument(
-        "--no-philter", action="store_false", dest="philter", default=True, help="Don’t run philter on notes"
+        "--no-philter", action="store_const", const=PHILTER_DISABLE, dest="philter", help=argparse.SUPPRESS
```
dogversioning, line 233: very kind of you, but i think you could make a breaking change now.

dogversioning (approved):

cumulus_etl/upload_notes/cli.py
mikix: Eh, I like to give some amount of gracetime, even to ourselves and our own muscle memory. But I've added a note that the arg can be removed after May 2024, to help avoid it staying around forever.

mikix:
OK tests fixed after adjusting to a `moto` API change: https://github.com/smart-on-fhir/cumulus-etl/pull/297/commits/90d2d51642cdb63e131edddd1773b8b1595c18c9

PR #299, opened 2024-03-13
Chart review doc gathering status

dogversioning:
This PR makes the following changes:
- Adds a status indicator for gathering documents
- Backs off the strict type enforcement for documents in favor of skipping and reporting an error.

### Checklist

### DIFF
--- a/cumulus_etl/common.py
+++ b/cumulus_etl/common.py
@@ -12,6 +12,7 @@ from collections.abc import Iterator
 from typing import Any, Protocol, TextIO
 
 import rich
+from rich import progress
 
 from cumulus_etl import store
 
@@ -317,6 +318,13 @@ def print_header(name: str | None = None) -> None:
     rich.get_console().rule()
     if name:
         print(name)
+
+def get_transient_progress()-> (progress.Progress):
+    progress_bar = progress.Progress(
+        progress.TextColumn("[progress.description]{task.description}"),
+        refresh_per_second=0.1)
+    return progress_bar
+    
 
 
 ###############################################################################
--- a/cumulus_etl/fhir/fhir_utils.py
+++ b/cumulus_etl/fhir/fhir_utils.py
@@ -201,10 +201,13 @@ async def get_docref_note(client: FhirClient, docref: dict) -> str:
     if best_attachment_index < 0:
         # We didn't find _any_ of our target text content types.
         # A content type isn't required by the spec with external URLs, so it's possible an unmarked link could be good.
-        # But let's optimistically enforce the need for a content type ourselves by bailing here.
-        # If we find a real-world need to be more permissive, we can change this later.
+        # We've found a real world case where we've found missing data preventing note download; for now,
+        # we are just notifying about it and returning an empty string object.
+        # If we find a actually need this data, we can change this later.
         # But note that if we do, we'll need to handle downloading Binary FHIR objects, in addition to arbitrary URLs.
-        raise ValueError("No textual mimetype found")
+        
+        print(f"Found unexpected mime type '{mimetype}', skipping.")
+        return ""
 
     note = await _get_docref_note_from_attachment(client, attachments[best_attachment_index])
 
--- a/cumulus_etl/upload_notes/selector.py
+++ b/cumulus_etl/upload_notes/selector.py
@@ -48,28 +48,42 @@ def _create_docref_filter(
 
 def _filter_real_docrefs(docrefs_csv: str, docrefs: Iterable[dict]) -> Iterator[dict]:
     """Keeps any docrefs that match the csv list"""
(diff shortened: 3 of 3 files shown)


cumulus_etl/fhir/fhir_utils.py
```diff
@@ -201,10 +201,13 @@ async def get_docref_note(client: FhirClient, docref: dict) -> str:
...
-        # But let's optimistically enforce the need for a content type ourselves by bailing here.
-        # If we find a real-world need to be more permissive, we can change this later.
+        # We've found a real world case where we've found missing data preventing note download; for now,
+        # we are just notifying about it and returning an empty string object.
+        # If we find a actually need this data, we can change this later.
         # But note that if we do, we'll need to handle downloading Binary FHIR objects, in addition to arbitrary URLs.
-        raise ValueError("No textual mimetype found")
+        
+        print(f"Found unexpected mime type '{mimetype}', skipping.")
+        return ""
```
dogversioning, line 210: the type in question that caused this error was `application/xml` - wanted to talk about this change before commiting to this approach.

cumulus_etl/common.py
```diff
@@ -318,6 +319,13 @@ def print_header(name: str | None = None) -> None:
     if name:
         print(name)
 
+def get_transient_progress()-> (progress.Progress):
```
mikix, line 322: nit: this might more naturally fit in `cli_utils.py` (which is where `make_progress_bar()` lives). I'm trying to _roughly_ group misc functions. (Maybe even convert this into a `transient=False` kwarg on that method?)

How do you feel about whether a `TimeElapsed` column fits here? Helpful or noisy?

Also, why the parens on the return type hint? Is that something your IDE or `ruff` encourages?

cumulus_etl/fhir/fhir_utils.py
mikix: We can talk about this during our meeting today, but I'm guessing this was a Cerner thing they do - they wrap the doc in a Cerner-specific xml format, with an RDF doc base64-encoded inside. Usually they generate the html version of the doc later in the doc lifecycle, like when it becomes a `status=final` or whatever the status is.

So I'm leery of trying to support `application/xml` - we already (very optimistically) support `application/xhtml+xml`. But arbitrary xml is harder to know what to do with (i.e. how to extract the note text).

cumulus_etl/upload_notes/selector.py
```diff
@@ -48,28 +48,42 @@ def _create_docref_filter(
...
+        for docref in docrefs:
+            if docref["id"] in real_docref_ids:
+                found += 1
+                yield docref
+
+                real_docref_ids.remove(docref["id"])
+                if not real_docref_ids:
+                    break
+            scanned += 1
+            progress.update(task, description = f"Unmatched IDs: {len(real_docref_ids)}, found: {found}, scanned: {scanned}")
```
mikix, line 68: minor nit: you'll miss the last progress update here if we early exit because we found all the target docrefs.

cumulus_etl/upload_notes/selector.py
```diff
@@ -48,28 +48,42 @@ def _create_docref_filter(
...
 
 def _filter_fake_docrefs(codebook: deid.Codebook, anon_docrefs_csv: str, docrefs: Iterable[dict]) -> Iterator[dict]:
```
mikix, line 70: These two filter methods were already very similar (and maybe should have been sharing more code between them). With the extra progress reporting, this is triggering my code sense that they should share more of their guts.

The only real difference is how to calculate the id to check for each docref, right? That could be a lambda arg or something, I dunno.

cumulus_etl/fhir/fhir_utils.py
mikix: OK after talking about this in person, I think your beef was more that unhandled mimetypes were stopping `upload-notes` in its tracks, rather than you wanting to handle application/xml. That makes sense.

So I might vote for `upload-notes` catching the ValueError here and printing something sensible there. This method is also used by the ETL and I believe it catches errors and prints the message itself. So that's the established error-processing approach for this anyway.

dogversioning: I'm fine with that - is it cool if I keep the erroring mimetype info in the ValueError?

cumulus_etl/common.py
dogversioning: oh at one point i was passing back a tuple - i'll remove those.

TimeElapsed is a fine addition.

cumulus_etl/fhir/fhir_utils.py
mikix: Sure - I was having vibes that it might be too low level info, but :shrug: give the people what they crave - mimetypes

Issue #298, opened 2024-03-12
include DATE in heading for import to labelstudio


comorbidity:
Please include DATE in the header for each section. 

For example

########################################
########################################
Neurosurgery Visit 
2024/03/12
########################################
########################################

mikix:
Yeah that seems reasonable! A DocRef might not have the `date` field (though US Core does say it's a "must support" field, so hopefully it's filled out). But we can note that like `Unknown date` or something, for that rare case.

mikix:
I don't think we currently sort the DocRefs either? We should probably sort by `date` too while we're inspecting that field.

PR #300, opened 2024-03-14
feat(upload-notes): add datetimes to each DocRef header

mikix:
Also:
- Order the DocRefs in an encounter by datetime
- Switch from freezegun to time-machine

Unrelatedly:
- Switch from deprecated cgi module to email module for mimetype parsing.

Fixes https://github.com/smart-on-fhir/cumulus-etl/issues/298

### Checklist

### DIFF
--- a/cumulus_etl/fhir/fhir_utils.py
+++ b/cumulus_etl/fhir/fhir_utils.py
@@ -1,7 +1,8 @@
 """FHIR utility methods"""
 
 import base64
-import cgi
+import datetime
+import email.message
 import re
 
 import inscriptis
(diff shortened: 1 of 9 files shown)


cumulus_etl/fhir/fhir_utils.py
```diff
@@ -104,9 +142,9 @@ async def download_reference(client: FhirClient, reference: str) -> dict | None:
 
 def _parse_content_type(content_type: str) -> (str, str):
     """Returns (mimetype, encoding)"""
-    # TODO: switch to message.Message parsing, since cgi is deprecated
-    mimetype, params = cgi.parse_header(content_type)
-    return mimetype, params.get("charset", "utf8")
+    msg = email.message.EmailMessage()
+    msg["content-type"] = content_type
+    return msg.get_content_type(), msg.get_content_charset("utf8")
```
mikix, line 147: While I was here, I noticed this and figured I'd finally switch this up.

cumulus_etl/upload_notes/cli.py
```diff
@@ -53,6 +54,15 @@ async def gather_docrefs(
         )
 
 
+def datetime_from_docref(docref: dict) -> datetime.datetime | None:
+    """Returns the date of a docref - preferring `context.period.start`, then `date`"""
```
mikix, line 58: This is preferring clinically-relevant time over the time-of-DocRef creation. That feels right to me, but I'm open to that being flipped.

tests/utils.py
```diff
@@ -13,21 +13,20 @@
...
+# Pass a non-UTC time to time-machine to help notice any bad timezone handling.
 # But only bother exposing the UTC version to other test code, since that's what will be most useful/common.
 _FROZEN_TIME = datetime.datetime(2021, 9, 15, 1, 23, 45, tzinfo=datetime.timezone(datetime.timedelta(hours=4)))
 FROZEN_TIME_UTC = _FROZEN_TIME.astimezone(datetime.timezone.utc)
 
 
 # Several tests involve timestamps in some form, so just pick a standard time for all tests.
-# We ignore socketserver because it checks the result of time() when evaluating timeouts.
-@freezegun.freeze_time(_FROZEN_TIME, ignore=["socketserver"])
+@time_machine.travel(_FROZEN_TIME, tick=False)
```
mikix, line 29: This switch fixes a crash I was seeing between pandas and freezegun (pandas only likes to use real DateTime objects, not fake ones and freezegun was generating mocked objects). Time Machine instead changes things at a lower level, and pandas seems happy.

It also fixes the issue with `socketserver` magically (despite time not ticking forward.... :shrug:)

cumulus_etl/upload_notes/cli.py
dogversioning: I think that is probably the right call.

cumulus_etl/upload_notes/cli.py
```diff
@@ -53,6 +54,15 @@ async def gather_docrefs(
...
+    if start := fhir.parse_datetime(docref.get("context", {}).get("period", {}).get("start")):
+        return start
+    if date := fhir.parse_datetime(docref.get("date")):
+        return date
```
dogversioning, line 62: aw yeah there's our new friends

tests/utils.py
dogversioning: oh that's interesting - i've had to jump in hoops around that, i'll look into it.

dogversioning (approved):

PR #303, opened 2024-03-18
Print better error "bad label studio config" message and update black

mikix:
I know we're also experimenting with ruff, but before committing to that, this is an easy update.

### Checklist

### DIFF
--- a/cumulus_etl/errors.py
+++ b/cumulus_etl/errors.py
@@ -3,6 +3,8 @@
 import sys
 from typing import NoReturn
 
+import rich.console
+
 
 # Error return codes, mostly just distinguished for the benefit of tests.
 # These start at 10 just to leave some room for future use.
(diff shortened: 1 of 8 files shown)


cumulus_etl/errors.py
```diff
@@ -56,5 +58,5 @@ class FatalError(Exception):
 
 def fatal(message: str, status: int) -> NoReturn:
     """Convenience method to exit the program with a user-friendly error message a test-friendly status code"""
-    print(message, file=sys.stderr)
+    rich.console.Console(stderr=True).print(message, style="bold red", highlight=False)
```
mikix, line 61: I tested this red config (which I stole from an example in `rich`'s docs on printing to the error console) on several different terminal color schemes -- seemed decent to me. Here's an example:

Screenshot from 2024-03-18 15-29-56

dogversioning (approved):

Issue #296, opened 2024-02-09
Mark which resources are loaded for which patients (i.e. "completion tracking")


mikix:
This comes from a study need:
- Sometimes loading data from the EHR can take a long time and/or can happen in fits and starts. You may take weeks to fully finish loading a set of cohorts.
- During that time, studies (including the `core__` tables) would probably want to ignore patients that don't have the resources they care about loaded.
- i.e. _studies want to be able to know whether the Conditions table is accurate for patient X or not - and exclude the patient if not._

My initial thoughts on this are to have the ETL keep a metadata table around, marking which resources are "finished" at the Group level. And then which patients belong to which Groups. That way a study could ask if patient X has Conditions yet.

Brainstorming for that approach:
- UX for ETL:
  - --auto-mark (looks for log file adjacent to input files, or from URL if ETL is doing export)
  - --mark GROUP_ID (alternative option if user wants to override)
  - If no mark detected or given, all this logic below will be skipped
  - Not in love with "mark"... --mark-complete --mark-finished --complete --finish...
- ETLing a resource group:
  - writes to a table like etl_complete
  - with columns (GROUP_ID, RESOURCE_NAME, ETL_DATE, NEWEST_LAST_UPDATED_DATE, OLDEST_LAST_UPDATED_DATE)
  - ETL will add new row when it successfully finishes a run
  - Open question: should I update a single unique group_id/resource_name row instead of appending to table? Might make querying less messy? But at cost of some potentially useful log records.
- ETLing ~patients~ encounters:
  - When uploading ~patients~ encounters, ETL will also write all row IDs to a table like ~etl_patient_groups~ etl_encounter_groups
  - with columns (~PATIENT_ID~ ENCOUNTER_ID, GROUP_ID)
  - Non-unique, as resources can be in multiple groups
- Library can tell all resources we have uploaded at least once for a given encounter by doing something like:
```sql
    SELECT DISTINCT etl_complete.resource_name
    FROM etl_encounter_groups
    INNER JOIN etl_complete
    ON etl_complete.group_id = etl_encounter_groups.group_id
    WHERE etl_encounter_groups.encounter_id = 'xxx'
```
- Assumptions of this approach:
[...]

PR #304, opened 2024-03-26
feat: initial completeness-tracking work

mikix:
This is early support for completeness-tracking (the ability to mark which groups & resources have been loaded by the ETL and are thus ready for studies to use).

- Adds some new (secret) CLI arguments: `--export-group`, `--export-timestamp`, and `--write-completion`
- Adds a new `etl__completion` table which holds:
  - table
  - group
  - export_time
- Adds a new `etl__completion_encounters` table which holds:
  - encounter_id
  - group
  - export_time
- This table is automatically written to, using the CLI values
- Currently, those arguments are optional. A future change will make them required. (though hopefully usually automatically inferred from export logs)
- When using the ndjson output format, you can no longer have any files in the output folder. This is to safeguard against accidents (and to make some code paths simpler)

This PR starts to address https://github.com/smart-on-fhir/cumulus-etl/issues/296 (Library side [here](https://github.com/smart-on-fhir/cumulus-library/pull/203))

**Future Plans**
This PR is just an preparatory, disabled-by-default step 1. These are bits of work that still have to be done to make this feature real (and required) across Cumulus.
- Parse export logs to autodetect arg values
- Require the new ETL args (from user, from log, or from our own export)
- Prevent overwriting new group data with old, to solve Epic use cases (check dates before deltalake write)
- Generate export logs from our own bulk export implementation, allowing us to parse it back later
- Backfill/migration support (maybe as simple as running `--group-name legacy --timestamp 2020` for all current encounters)
- Handle empty input sets - may require patching the bulk exporter to leave an empty file in those cases
- User docs

### Checklist

### DIFF
--- a/cumulus_etl/loaders/fhir/bulk_export.py
+++ b/cumulus_etl/loaders/fhir/bulk_export.py
@@ -1,6 +1,7 @@
 """Support for FHIR bulk exports"""
 
 import asyncio
+import datetime
 import json
 import os
 import urllib.parse
@@ -54,6 +55,20 @@ class BulkExporter:
         self._since = since
         self._until = until
 
+        # Public properties, to be read after the export:
+        self.export_datetime = None
+
+        # Parse the group out of the URL, which will look something like:
+        # - https://hostname/root/Group/my-group$export  <- group name of `my-group`
+        # - https://hostname/root$export <- no group name, global export
+        if "/Group/" in self._url:
+            latter_half = self._url.split("/Group/", 2)[-1]
+            self.group_name = latter_half.split("/")[0]
+        else:
+            # Global exports don't seem realistic, but the user does do them,
+            # we'll use the empty string as the default group name for that...
+            self.group_name = ""
+
     async def export(self) -> None:
         """
         Bulk export resources from a FHIR server into local ndjson files.
@@ -99,6 +114,8 @@ class BulkExporter:
 
             # Finished! We're done waiting and can download all the files
             response_json = response.json()
+
+            self.export_datetime = datetime.datetime.fromisoformat(response_json["transactionTime"])
 
             # Were there any server-side errors during the export?
             # The spec acknowledges that "error" is perhaps misleading for an array that can contain info messages.
--- a/tests/etl/base.py
+++ b/tests/etl/base.py
@@ -1,5 +1,6 @@
 """Base classes for ETL-oriented tests"""
 
+import datetime
 import os
 import shutil
 import tempfile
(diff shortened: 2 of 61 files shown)


cumulus_etl/loaders/fhir/bulk_export.py
```diff
@@ -54,6 +55,17 @@ def __init__(
...
+        # Public properties, to be read after the export:
+        self.export_datetime = None
+
+        if "/Group/" in self._url:
+            latter_half = self._url.split("/Group/", 2)[-1]
+            self.group_name = latter_half.split("/")[0]
+        else:
+            # Global exports don't seem realistic, but the user does do them,
+            # we'll use the empty string as the default group name for that...
+            self.group_name = ""
```
dogversioning, line 67: suggestion: maybe add some examples of expected _url values?

tests/etl/base.py
```diff
@@ -31,12 +32,12 @@ def setUp(self):
         self.root_path = os.path.join(self.datadir, self.DATA_ROOT)
         self.input_path = os.path.join(self.root_path, "input")
 
-        tmpdir = tempfile.mkdtemp()
+        self.tmpdir = tempfile.mkdtemp()
         # Comment out this next line when debugging, to persist directory
-        self.addCleanup(shutil.rmtree, tmpdir)
+        # self.addCleanup(shutil.rmtree, self.tmpdir)
```
dogversioning, line 37: should this be uncommented again?

tests/etl/test_tasks.py
```diff
@@ -123,6 +123,159 @@ async def test_batch_is_given_schema(self):
...
 
+@ddt.ddt
+class TestTaskCompletion(TaskTestCase):
+    """Tests for etl__completion* handling"""
+
+    async def test_encounter_completion(self):
+        """Verify that we write out completion data correctly"""
+        self.make_json("Encounter.1", "A")
+        self.make_json("Encounter.2", "B")
+        self.make_json("Encounter.3", "C")
```
dogversioning, line 134: nit: this may just be me, but generally i've been finding it hard to keep track of the context in my head when tests have IDs/Groups as a single character what the string is actually supposed to represent. Not saying change it, just something to consider.

dogversioning (approved):

cumulus_etl/completion/output.py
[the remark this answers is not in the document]
mikix: It's mostly schema stuff... `schema.py` feels a bit better.

tests/etl/base.py
mikix: Yes, thx

tests/etl/test_tasks.py
mikix: That's totes fair... I was using single characters for width savings, but it might not be worth the lower readability. I'll make a little pass and try to add a few more descriptive names.

mikix: I renamed a few variables - and will try to keep an eye on myself in the future :smile:

PR #306, opened 2024-04-12
feat: use group-name/export-time from log file if present

mikix:
When loading ndjson from a folder, parse any sibling logs we find for export info like group/time.

These are the current workflow assumptions this adds (but this is all still optional - no error will happen if a log doesn't exist):

- There will be a log.ndjson or log.*.ndjson file in the given folder.
  - There cannot be multiples (unless log.ndjson exists, in which case we always use that)
- That log file will be for a single export.
  - e.g. We will generally grab the last "kickoff" event and ignore others.

### Checklist


cumulus_etl/loaders/fhir/export_log.py
```diff
@@ -0,0 +1,99 @@
...
+        # log files with maybe a false start at the beginning).
+        try:
+            for row in common.read_ndjson(path):
+                match row.get("eventId"):
+                    case "kickoff":
+                        self._parse_kickoff(row)
+                    case "status_complete":
+                        self._parse_status_complete(row)
+        except (KeyError, json.JSONDecodeError) as exc:
+            raise self.IncompleteLog(f"Error parsing '{path}'") from exc
```
dogversioning, line 57: suggestion: consider a MissingLog error?

dogversioning (approved):

PR #308, opened 2024-05-06
feat: write out a bulk export log file when we do the export

mikix:
When Cumulus ETL is running the bulk export ourselves, we now write out a bulk export log, in the standard[1] format.

This is only useful when passing an --export-to folder of course, but that ought to be the normal use case.

Also:
- Don't early exit an export if there are server-provided error messages. We'll still bail, because our exporter is very conservative. But we'll do that at the end, after we download everything. This should help with debugging issues and archiving.

[1] https://github.com/smart-on-fhir/bulk-data-client/wiki/Bulk-Data-Export-Log-Items

### Checklist

### DIFF
--- a/cumulus_etl/__init__.py
+++ b/cumulus_etl/__init__.py
@@ -1,3 +1,3 @@
 """Cumulus public entry point"""
 
-__version__ = "1.0.0"
+__version__ = "1.1.0"
--- a/cumulus_etl/loaders/fhir/export_log.py
+++ b/cumulus_etl/loaders/fhir/export_log.py
@@ -8,8 +8,11 @@ import datetime
 import json
 import os
 import re
+import uuid
 
-from cumulus_etl import common, fhir, store
+import httpx
+
+from cumulus_etl import common, errors, fhir, store
 
 
 class BulkExportLogParser:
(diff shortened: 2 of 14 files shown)


cumulus_etl/__init__.py
```diff
@@ -1,3 +1,3 @@
 """Cumulus public entry point"""
 
-__version__ = "1.0.0"
+__version__ = "1.1.0"
```
mikix, line 3: Finally trying to be more disciplined about version bumps, even though we don't do pip releases

docs/setup/cumulus-aws-template.yaml
```diff
@@ -198,8 +198,6 @@ Resources:
               - !Sub "s3://${S3Bucket}/${EtlSubdir}/servicerequest"
               - !Sub "s3://${S3Bucket}/${EtlSubdir}/covid_symptom__nlp_results"
               - !Sub "s3://${S3Bucket}/${EtlSubdir}/covid_symptom__nlp_results_term_exists"
-              - !Sub "s3://${S3Bucket}/${EtlSubdir}/etl__completion"
-              - !Sub "s3://${S3Bucket}/${EtlSubdir}/etl__completion_encounters"
```
mikix, line 202: I just did this so that folks don't reference these tables yet until we're 100% ready - if a crawler tries to crawl a dir that doesn't exist, and then you query what the metadata for that table is (as Library does, to see what it can use), you get errors. :frowning:

pyproject.toml
```diff
@@ -20,7 +20,7 @@ dependencies = [
     "label-studio-sdk < 1",
     "oracledb < 3",
     "philter-lite < 1",
-    "pyarrow < 16",
+    "pyarrow < 17",
```
mikix, line 23: I don't remember what changed now, but minor deprecation stuff.

docs/setup/cumulus-aws-template.yaml
dogversioning: do you want to comment them out instead?

cumulus_etl/loaders/fhir/export_log.py
```diff
@@ -97,3 +100,186 @@ def _find(self, root: store.Root) -> str:
...
+
+    def _event(self, event_id: str, detail: dict, *, timestamp: datetime.datetime = None) -> None:
+        timestamp = timestamp or common.datetime_now(local=True)
+        if self._start_time is None:
+            self._start_time = timestamp
+
+        # We open the file anew for each event because:
+        # a) logging should be flushed often to disk
+        # b) it makes the API of the class easier by avoiding a context manager
+        with self.root.fs.open(self._filename, "a", encoding="utf8") as f:
```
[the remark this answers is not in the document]
mikix: We currently require `--export-to` folders to be local, FYI. (Largely because the MS deid tool needs a local dir, and we were operating directly on the export dir.)

With this PR, where we now copy the export data to a local folder for the MS tool (because the MS tool would choke on the log file), we _could_ explore relaxing that requirement.

And thus, the log file could be on an S3 bucket. This code here is already using the `fsspec` API, so it could write to S3 (if it transparently supports the "append" write mode...?)

I will say, for my own purposes, I like to export locally and inspect the results for sanity & a lack of errors before "promoting" the results to S3. But maybe that's me treating S3 as special, when we could just write results to it, and delete them if they are bad.

docs/setup/cumulus-aws-template.yaml
mikix: If this were code for devs, I'd be down for that - but since this is like, public template to be copied, I just want to avoid any questions about what these commented lines might mean until I'm ready to announce how to use this.

I probably should not have even added these here, before I was ready to go public.

PR #310, opened 2024-05-09
docker: depend on ctakes 1.1.1

mikix:
This hopefully fixes a race condition when restarting cTAKES as we feed it new bsv files.

Also bumped ETL's version to 1.1.1 (coincidentally the same :D)

### Checklist

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -37,7 +37,7 @@ services:
       - etl-gpu
 
   ctakes-covid-base:
-    image: smartonfhir/ctakes-covid:1.1.0
+    image: smartonfhir/ctakes-covid:1.1.1
     environment:
       - ctakes_umlsuser=umls_api_key
       - ctakes_umlspw=$UMLS_API_KEY
--- a/cumulus_etl/__init__.py
+++ b/cumulus_etl/__init__.py
@@ -1,3 +1,3 @@
 """Cumulus public entry point"""
 
-__version__ = "1.1.0"
+__version__ = "1.1.1"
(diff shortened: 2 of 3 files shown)


compose.yaml
```diff
@@ -37,7 +37,7 @@ services:
       - etl-gpu
 
   ctakes-covid-base:
-    image: smartonfhir/ctakes-covid:1.1.0
+    image: smartonfhir/ctakes-covid:1.1.1
```
mikix, line 40: This is very tight pinning, I know. And maybe it should just be `1.1`? But I also kind of like how paranoid it is. Open to thoughts.

cumulus_etl/__init__.py
```diff
@@ -1,3 +1,3 @@
 """Cumulus public entry point"""
 
-__version__ = "1.1.0"
+__version__ = "1.1.1"
```
mikix, line 3: Just a coincidence that this lined up like this, I didn't just do a search and replace :smile:

compose.yaml
dogversioning: i'm ok in this case since its unlikely to change, and we always want to run latest.

dogversioning (approved):

compose.yaml
mikix: You're OK with what? `1.1` or the paranoia?

While we always want to run the latest, if there's every any change that can affect the results from cTAKES, we need to bump some task version numbers in every task that uses cTAKES.

So like, I wouldn't want to use just `1` or `latest` as the tag. And I went extra paranoid by also wanting to consider every patch release for possible changes. Which is maybe too much, since we are co-maintainers of that docker image and can enforce semantic versioning. But only co-maintainers...

mikix: (we need to bump task version numbers to (a) reset any existing ctakes caches you may have and (b) flag new task results so you can track which bits need to be re-run with the new NLP changes)

PR #314, opened 2024-05-20
fix: be more graceful when MS tool moves files underneath us

mikix:
It moves temporary files into their final place, and that might happen as we are trying to get the file size - so handle that exception.

### Checklist

### DIFF
--- a/cumulus_etl/deid/mstool.py
+++ b/cumulus_etl/deid/mstool.py
@@ -80,6 +80,17 @@ def _compare_file_sizes(target: dict[str, int], current: dict[str, int]) -> floa
     return total_current / total_expected
 
 
+def _get_file_size_safe(path: str) -> int:
+    try:
+        return os.path.getsize(path)
+    except FileNotFoundError:
+        # The MS Tool moves temporary files around as it completes each file,
+        # so we guard against an unlucky race condition of a file being moved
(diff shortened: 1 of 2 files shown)


cumulus_etl/deid/mstool.py
```diff
@@ -80,6 +80,17 @@ def _compare_file_sizes(target: dict[str, int], current: dict[str, int]) -> floa
     return total_current / total_expected
 
 
+def _get_file_size_safe(path: str) -> int:
+    try:
+        return os.path.getsize(path)
+    except FileNotFoundError:
```
Dtphelan1, line 86: (comment being written from an outsider's perspective, so forgive stupid questions plz)
Is this the right Error? The 3.12 docs I was looking at for `os.path` [suggest it would throw OSError](https://docs.python.org/3/library/os.path.html#os.path.getsize)

Dtphelan1 (approved):
Only one thing I double checked and wanted to be sure about, but otherwise I think this looks good to me. Once the Error thrown is either confirmed as correct here or fixed, then you're good to go

cumulus_etl/deid/mstool.py
mikix: Ah yeah as we mentioned on slack, it's a builtin subclass of OSError.

PR #316, opened 2024-05-21
Enable liquid clustering for delta lakes

mikix:
In practice, for most of our big raw tables, this will involve clustering on a single column: `id`, in which case Delta Lake falls back to a very simple ZORDER algorithm (making sure that the data files are ordered by `id`, to make looking up via `id` easy) -- which is still good, just not as sleek and cool as the full "liquid clustering" algorithm.

But it's still good to do for two reasons:
- This new paradigm is apparently the future of Delta Lake optimization, replacing normal ZORDER and partitioning. And you can't convert an existing table to it. So if we start creating clustered tables now, even if it's secretly just a ZORDER, we can benefit from future improvements to the algorithm and/or add more clustering columns if we want to.
- There are a few tables that *do* cluster by multiple columns and thus benefit from the cool liquid clustering logic - namely the completion tables. Those tables don't get crazy large anyway, but still - one of them scales with the number of encounters, so that's not chump change.

Delta Lake Protocol Notes:
- This bumps the delta lake writer protocol to 7 but keeps the reader protocol at 1 - this means Athena can keep reading these tables (I tested that). But it does mean that _very_ old versions of the ETL might not be able to write to these tables. Since we don't expect multiple ETL installations to be pointing at the same delta lakes often (usually just one engineer is doing that), because you have to create a table first before this becomes a problem, and because you'd need a truly ancient ETL, I'm not worried about this.

### Checklist

### DIFF
--- a/cumulus_etl/formats/deltalake.py
+++ b/cumulus_etl/formats/deltalake.py
@@ -101,6 +101,7 @@ class DeltaLakeFormat(Format):
         table = (
             delta.DeltaTable.createIfNotExists(self.spark)
             .addColumns(updates.schema)
+            .clusterBy(*self.uniqueness_fields)
             .location(self._table_path(self.dbname))
             .execute()
         )
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ requires-python = ">= 3.10"
 dependencies = [
     "ctakesclient >= 5.1, < 6",
     "cumulus-fhir-support >= 1.2, < 2",
-    "delta-spark >= 3, < 4",
+    "delta-spark >= 3.2.1, < 4",
     "httpx < 1",
     "inscriptis < 3",
     "jwcrypto < 2",
--- a/cumulus_etl/__init__.py
+++ b/cumulus_etl/__init__.py
@@ -1,3 +1,3 @@
 """Turns FHIR data into de-identified & aggregated records"""
 
-__version__ = "1.3.0"
+__version__ = "1.4.0"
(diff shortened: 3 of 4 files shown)


cumulus_etl/formats/deltalake.py
```diff
@@ -93,38 +93,33 @@ def _write_one_batch(self, batch: Batch) -> None:
...
-            # Determine merge condition
-            conditions = [f"table.{field} = updates.{field}" for field in self.uniqueness_fields]
-            condition = " AND ".join(conditions)
+        table = (
+            delta.DeltaTable.createIfNotExists(self.spark)
+            .addColumns(updates.schema)
+            .clusterBy(*self.uniqueness_fields)
+            .location(self._table_path(self.dbname))
+            .execute()
+        )
```
mikix, line 102: This is the interesting new bit. The rest of the delta is unindenting and removing the except handler that created the table.

Now, we just create/get the table in the same go. Using the delta lake specific method was the only way I could find to specify clustering - and it's a cleaner python code path.

pyproject.toml
```diff
@@ -13,7 +13,7 @@ requires-python = ">= 3.10"
 dependencies = [
     "ctakesclient >= 5.1, < 6",
     "cumulus-fhir-support >= 1.2, < 2",
-    "delta-spark >= 3, < 4",
+    "delta-spark >= 3.2.1, < 4",
```
mikix, line 16: 3.2.1 had a specific fix for "clustering by a single column" that we were waiting for, since that's our primary use case here

dogversioning (approved):

PR #321, opened 2024-06-07
fix(upload-notes): allow uploading to dynamic label setups again

mikix:
Newer versions of Label Studio seem to flag dynamic labels differently than previous ones. So detect dynamic labels a different way.

Also:
- Print how many notes we are attempting to upload.
- Warn if we don't see any DocumentReference files.
- Add "-f compose.yaml" to the upload-notes example commands in the docs as a reminder that you need to be running off a compose file.

### Checklist

### DIFF
--- a/cumulus_etl/upload_notes/labelstudio.py
+++ b/cumulus_etl/upload_notes/labelstudio.py
@@ -193,7 +193,16 @@ class LabelStudioClient:
         }
 
     def _update_used_labels(self, task: dict, used_labels: Iterable[str]) -> None:
+        uses_dynamic_labels = False
         if self._labels_config.get("dynamic_labels"):
+            # Old versions of Label Studio set this nice field for us.
+            uses_dynamic_labels = True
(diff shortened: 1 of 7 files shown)


cumulus_etl/upload_notes/labelstudio.py
```diff
@@ -193,7 +193,16 @@ def _format_philter_span(self, count: int, start: int, end: int, note: LabelStud
         }
 
     def _update_used_labels(self, task: dict, used_labels: Iterable[str]) -> None:
+        uses_dynamic_labels = False
         if self._labels_config.get("dynamic_labels"):
+            # Old versions of Label Studio set this nice field for us.
+            uses_dynamic_labels = True
+        elif self._labels_config.get("labels") == []:
+            # Newer versions of Label Studio seem to just pass an empty labels list.
+            # (An empty list would not normally be allowed in a static setup.)
```
dogversioning, line 202: 😬

dogversioning (approved):

PR #326, opened 2024-07-02
docs: add completion tables in our AWS template

mikix:
These aren't necessarily used, but it's lightly helpful to have them already in place if you want to experiment with the completion tracking feature (--write-completion)

### Checklist

### DIFF
--- a/docs/setup/cumulus-aws-template.yaml
+++ b/docs/setup/cumulus-aws-template.yaml
@@ -198,6 +198,8 @@ Resources:
               - !Sub "s3://${S3Bucket}/${EtlSubdir}/servicerequest"
               - !Sub "s3://${S3Bucket}/${EtlSubdir}/covid_symptom__nlp_results"
               - !Sub "s3://${S3Bucket}/${EtlSubdir}/covid_symptom__nlp_results_term_exists"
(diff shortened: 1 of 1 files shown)


docs/setup/cumulus-aws-template.yaml
```diff
@@ -198,6 +198,8 @@ Resources:
               - !Sub "s3://${S3Bucket}/${EtlSubdir}/servicerequest"
               - !Sub "s3://${S3Bucket}/${EtlSubdir}/covid_symptom__nlp_results"
               - !Sub "s3://${S3Bucket}/${EtlSubdir}/covid_symptom__nlp_results_term_exists"
+              - !Sub "s3://${S3Bucket}/${EtlSubdir}/etl__completion"
+              - !Sub "s3://${S3Bucket}/${EtlSubdir}/etl__completion_encounters"
```
mikix, line 202: Yes I removed these lines in the past to not confuse people - but that was because an empty crawler table was causing Library errors - I fixed that in https://github.com/smart-on-fhir/cumulus-library/pull/236 so these should be safe to add back in.

dogversioning (approved):

Issue #135, opened 2023-01-05
Support HTTP proxies for web requests


dogversioning:
Pretty standard territory on this one - https://www.scrapingbee.com/blog/python-requests-proxy/ , among other resources explaining the same - we just need to decide how we want to inject it (perhaps a .ENV file).

PR #329, opened 2024-07-10
feat: pass common networking variables through the docker layer

### LINKED ISSUE #319: Allow specifying a custom CA root for https
An org might have a custom CA setup for https signing.

We can [pass httpx a new one](https://www.python-httpx.org/advanced/ssl/) via `verify=`. And grab that location via env and/or CLI arg.

Or maybe provide docs on mounting a file inside the container in an appropriate place for linux? Not sure how customizable httpx is in that regard - where on disk it looks and all that.

mikix:
This should allow http proxies and custom CA roots.

Fixes #319 and I think #135 too.

### Checklist

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -1,6 +1,26 @@
 services:
 
+  # This service is shared by all other services - be careful when adding to it.
+  common-base:
+    environment:
+      # These are commonly-supported networking environment variables.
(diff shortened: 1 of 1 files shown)


dogversioning (approved):
proxies! what a mistake!

PR #331, opened 2024-07-29
Switch the ETL to ruff

mikix:
- Switch to ruff (first commit)
- Meet Matt halfway at 100 line width (second commit)
- Dropped pylint disable lines (third commit)

Interesting deltas with Library's ruff config:
- I dropped the ISC checks (only 3 of them and 2 of them actively conflict with the formatter - I'm just letting `format` handle this)
- I dropped the B (bugbear) checks - they were marked as security checks, but that's not really what bugbear is doing. And I didn't love two of the changes it wanted ([B027](https://docs.astral.sh/ruff/rules/empty-method-without-abstract-decorator/) and [B905](https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/) -- 905 I can kind of see, but 027 especially I didn't love)
- I added S (bandit) checks to get back security scanning - I'd recommend the same in Library

### Checklist


dogversioning (commented):
You can keep the 120 if it makes you happy, you're the boss over here.

PR #332, opened 2024-07-31
feat: add new "cumulus-etl export" command for bulk exporting

mikix:
This new command is quite simple. It just takes a URL and an output folder. The usual auth and export args (like --since) are accepted.

You can also filter exported resources by task with --task.

But even if you don't use --since or --task, you can provide an export URL that has _since, _type, _typeFilter, whatever you want.

In addition:
- "--task=help" will now print a list of task names and exit.
- The traditional way to kick off a bulk export when doing an ETL job (by providing a URL as the input path) now also lets you specify custom parameters like _since or _typeFilter directly in the URL.
- Handle non-compliant servers that give us transactionTime values that aren't formatted correctly (as bulk-data-server does).

### Checklist

### DIFF
--- a/cumulus_etl/loaders/fhir/bulk_export.py
+++ b/cumulus_etl/loaders/fhir/bulk_export.py
@@ -40,6 +40,7 @@ class BulkExporter:
         destination: str,
         since: str | None = None,
         until: str | None = None,
+        prefer_url_resources: bool = False,
     ):
         """
         Initialize a bulk exporter (but does not start an export).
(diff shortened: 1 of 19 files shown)


tests/loaders/ndjson/test_bulk_export.py
```diff
@@ -240,7 +244,7 @@ async def test_happy_path(self):
     async def test_since_until(self):
         """Verify that we send since & until parameters correctly to the server"""
         self.mock_kickoff(
-            params="?_type=Condition%2CPatient&_since=2000-01-01T00%3A00%3A00%2B00.00&_until=2010",
+            params="?_type=Condition&_type=Patient&_since=2000-01-01T00%3A00%3A00%2B00.00&_until=2010",
```
mikix, line 247: Generating a URL with duplicate _type fields vs a comma-separated field is an oddity of exploding the URL and putting it back together with urlparse. I could have manually fixed it after the fact, but the [spec](https://hl7.org/fhir/uv/bulkdata/export.html#bulk-data-kick-off-request) seems to prefer duplicate fields (even threatening that if folks don't love the comma approach, they'll take it out -- which seems highly unlikely at this adoption point, but still, I thought it was an interesting note.

cumulus_etl/loaders/fhir/bulk_export.py
```diff
@@ -125,7 +149,11 @@ async def export(self) -> None:
         # Finished! We're done waiting and can download all the files
         response_json = response.json()
 
-        self.export_datetime = datetime.datetime.fromisoformat(response_json["transactionTime"])
+        try:
+            transaction_time = response_json.get("transactionTime", "")
+            self.export_datetime = datetime.datetime.fromisoformat(transaction_time)
+        except ValueError:
+            pass  # server gave us a bad timestamp, ignore it :shrug:
```
dogversioning, line 156: you could? take the system time, which would be pretty close?

tests/loaders/ndjson/test_bulk_export.py
dogversioning: that is very unusual.

dogversioning (approved):

cumulus_etl/loaders/fhir/bulk_export.py
mikix: Well, this isn't a super realistic event in the first place - spec is pretty clear about the format. But also... it's tricky. This timestamp needs to be precise. Spec says:
```
The response SHOULD NOT include any resources modified after this instant, and SHALL include any matching resources modified up to and including this instant. 
```

So if we created a backup timestamp, we couldn't make those same guarantees. I'd rather just not claim anything in this weird edge case. (This timestamp is really only used for the off-by-default completion-tracking feature, where precision is useful.)

PR #335, opened 2024-08-02
WIP: Add --resume arg to continue an interrupted bulk export

mikix:
### Checklist


cumulus_etl/cli_utils.py
```diff
@@ -45,9 +47,14 @@ def add_aws(parser: argparse.ArgumentParser) -> None:
...
-    parser.add_argument("--since", help="start date for export from the FHIR server")
+    parser.add_argument(
+        "--since", metavar="TIMESTAMP", help="start date for export from the FHIR server"
+    )
     # "Until" is not an official part of the bulk FHIR API, but some custom servers support it
-    parser.add_argument("--until", help="end date for export from the FHIR server")
+    parser.add_argument(
+        "--until", metavar="TIMESTAMP", help="end date for export from the FHIR server"
+    )
+    parser.add_argument("--resume", metavar="URL", help="polling status URL from a previous export")
```
mikix, line 57: Note that this flag is being added to both the standalone `cumulus-etl export` command and the default `cumulus-etl etl` command.

.github/workflows/ci.yaml
```diff
@@ -88,7 +88,9 @@ jobs:
           tags: smartonfhir/cumulus-etl:latest
 
       - name: Download NLP images
-        run: docker compose --profile covid-symptom up -d --quiet-pull
+        run: |
+          # Just start the specific services we need, to conserve CI disk space
+          docker compose up -d --quiet-pull cnlpt-negation ctakes-covid
```
mikix, line 93: Unrelated fix put in as a separate commit - CI was failing because the image was running out of space... GitHub gives us 14GB (but then of course the OS and other stuff that actions install like node etc take up a lot).

dogversioning:
I'm going to approve this separately from my comments on the PR description, since I think this approach is fine and my take is very subjective.

With that said, from my end:
- I don't know that I care about matching the bulk client's CLI UX
- We have all the information needed to give a user the exact command they should rerun, rather than just a param to add

So to my mind, we could spit out a command with the necessary resume arguments, and exclude all the extra args, and that would be fine - it's functionally going to be the same action (copy/paste some text), and will be a little more explicit on 'what are these two things actually doing when invoked'. But feel free to disagree.

I also don't hate the idea of writing out a a command or config dict to disk alongside the download, and pointing a user to the path to that file, and have the resume command just load in whatever you need from there, and delete that when the download completes successfully, though that does externalize the functionality in a way that's not terribly clean.

dogversioning (approved):

mikix:
OK we talked offline - we think we're "fine" with this approach as is (the `--resume=URL` approach). We don't think we're locking ourselves into this, if we want to change how this works in the future. We remain curious about some magic log-reading resume feature - could add in the future maybe. But for now, we'll just go with this (and add a section in the docs about it).

Issue #334, opened 2024-07-31
Make bulk exporting more robust to transient errors


mikix:
EHR servers are wild out there, giving random 500 or other errors sometimes.

We do have retry support already in there for 202 & 429 errors. We could extend that to some level of retrying (with limits) for other random errors.

This together with #333 would make the bulk exporting feature much more reliable.

PR #336, opened 2024-08-06
Retry bulk export requests that yield 5xx server errors

mikix:
Retry each request four times (for a total of five requests) in an exponential backoff of 1, 2, 4, and 8 minutes (total of 15 minutes).

This should hopefully help when dealing with flaky EHRs.

Fixes: #334

### Checklist

### DIFF
--- a/cumulus_etl/fhir/fhir_client.py
+++ b/cumulus_etl/fhir/fhir_client.py
@@ -91,12 +91,14 @@ class FhirClient:
         """
         Issues an HTTP request.
 
-        The default Accept type is application/fhir+json, but can be overridden by a provided header.
+        The default Accept type is application/fhir+json, but can be overridden by a provided
+        header.
 
-        This is a lightly modified version of FHIRServer._get(), but additionally supports streaming and
-        reauthorization.
+        This is a lightly modified version of FHIRServer._get(), but additionally supports
+        streaming and reauthorization.
 
-        Will raise a FatalError for an HTTP error, except for 429 which gets returned like a success code.
+        Will raise a FatalError for an HTTP error, except for 429 which gets returned like a
+        success code.
 
         :param method: HTTP method to issue
         :param path: relative path from the server root to request
@@ -169,16 +171,17 @@ class FhirClient:
         """
         return self._capabilities
 
-    ###################################################################################################################
+    #############################################################################################
     #
     # Helpers
     #
-    ###################################################################################################################
+    #############################################################################################
 
     async def _read_capabilities(self) -> None:
         """
-        Reads the server's CapabilityStatement and sets any properties as a result (like server/vendor type).
+        Reads the server's CapabilityStatement and sets any properties as a result.
 
+        Notably, this gathers the server/vendor type.
         This is expected to be called extremely early, right as the http session is opened.
         """
         if not self._server_root:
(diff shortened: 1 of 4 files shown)


cumulus_etl/fhir/fhir_client.py
```diff
@@ -259,8 +262,8 @@ def create_fhir_client_for_cli(
             # Use the input URL as the base URL. But note that it may not be the server root.
             # For example, it may be a Group export URL. Let's try to find the actual root.
             client_base_url = root_input.path
-            client_base_url = re.sub(r"/Patient/?$", "/", client_base_url)
-            client_base_url = re.sub(r"/Group/[^/]+/?$", "/", client_base_url)
+            client_base_url = re.sub(r"/Patient($|/.*)", "/", client_base_url)
+            client_base_url = re.sub(r"/Group/[^/]+($|/.*)", "/", client_base_url)
```
mikix, line 266: Unrelated, but I discovered that now that we support input URLs with arguments like `/$export?_type=xxx`, this regex that tries to find the root for authentication purposes needed to handle that better.

cumulus_etl/loaders/fhir/bulk_export.py
```diff
@@ -236,90 +233,133 @@ async def _delete_export(self, poll_url: str) -> None:
...
-                        delay = min(delay, 300)
-                    progress_msg = response.headers.get("X-Progress", "waiting…")
-                    formatted_total = common.human_time_offset(self._total_wait_time)
-                    formatted_delay = common.human_time_offset(delay)
-                    status_box.plain = f"{progress_msg} ({formatted_total} so far, waiting for {formatted_delay} more)"
+        # Set up error handling variables.
+        # These times are extremely generous - partly because we can afford to be
+        # as a long-running async task and partly because EHR servers seem prone to
+        # outages that clear up after a bit.
+        error_retry_minutes = [1, 2, 4, 8]  # and then raise
```
dogversioning, line 290: is there any value in allowing a user to specify the number of retries?

cumulus_etl/loaders/fhir/bulk_export.py
```diff
@@ -236,90 +233,133 @@ async def _delete_export(self, poll_url: str) -> None:
...
-            raise
+            # 202 == server is still working on it, 429 == server is busy,
+            # 5xx == server-side error. In all cases, we wait.
+            if response.status_code in [202, 429] or response.is_server_error:
+                if log_progress and not response.is_server_error:
+                    log_progress(response)
+
+                # Calculate how long to wait, with a basic exponential backoff for errors.
+                if num_errors:
+                    default_delay = error_retry_minutes[num_errors - 1] * 60
```
dogversioning, line 328: nit - should there be one unified backoff scheme? should it be in one place

dogversioning (approved):

cumulus_etl/loaders/fhir/bulk_export.py
mikix: So far, I've just put retrying on the "application" side of the http code - i.e. not directly in FhirClient, but rather in the BulkExporter code.

Partly because I wanted to give myself the freedom to be a bit inflexible with it for now. Like... I dunno, is the user really going to want to specify number of retries? Feels too nitty gritty - if after 15minutes, a server still has (recoverable) problems, I'd rather just hear about it and then add another hard-coded retry.

The other odd thing about this retry code is that it's (intentionally) so slow. Most "exponential backoff" code out there is like, "wait 2 seconds, then 4, then 20 seconds" and I'm out here waiting 8 minutes.

So when I was thinking of building flexibility into FhirClient for number of retries and how long, just to use it in one place with weird values, I figured - just build the inflexible version first.

If we want retrying elsewhere in the ETL later, we can push this down a layer into FhirClient, with more flexibility. Does that sound OK?

dogversioning: yeah that's perfectly reasonable

Issue #341, opened 2024-08-28
Only show the error "can't download medication, need --fhir-url" once


mikix:
If a user forgot they needed that, we don't need to flood their console with that message, crowding out other possible error messages.

PR #342, opened 2024-08-28
medicationrequest: don't spam console with "need --fhir-url" errors

mikix:
If an error occurs while downloading Medication references, it's likely going to repeat itself (missing --fhir-url, server being down, etc).

So just print it once.

Fixes: #341 

### Checklist


Dockerfile
mikix, line 1: Just fixing the github warnings in PRs on this file.

dogversioning (approved):

PR #347, opened 2024-09-10
feat: don't simply toss extensions

### LINKED ISSUE #346: Print extensions and modifierExtensions before stripping them
This way a user can still get a sense of what extensions are in use / what they are missing. And then they can investigate further / submit a ticket for support for the extension as needed.

Maybe just print a number of resources with the extension, like so:
```
  25 Patients had extension: https://example.com/new-thing
```

mikix:
- Print a report at the end of the ETL run of all extensions stripped by resource, URL, and count.
- Plus a report on resources skipped due to unrecognized modifier extensions.
- Leave stub extension in place, with just the `url` field, so that we can inspect stripped extensions in SQL if we want.

This commit allows a bunch of extensions in use today. Scanned some sample data from Cerner and Epic to find common extensions. We'll surely add to the list going forward, to avoid annoying folks with the unknown-extension warning.

Ignore the following extensions:
- http://hl7.org/fhir/StructureDefinition/iso21090-TEL-address
- http://hl7.org/fhir/StructureDefinition/rendered-value
- http://hl7.org/fhir/us/core/StructureDefinition/us-core-direct
- https://fhir-ehr.cerner.com/r4/StructureDefinition/clinical-instruction
- https://fhir-ehr.cerner.com/r4/StructureDefinition/estimated-financial-responsibility-amount
- http://open.epic.com/FHIR/StructureDefinition/extension/birth-location

Drop support for the following extension:
- http://hl7.org/fhir/us/core/StructureDefinition/us-core-sex-for-clinical-use (this was in US Core 6.0 ballot, but didn't make final cut - it seems to be replaced by just 'us-core-sex')

Fixes #346

### Example output
```
────────────────────────────────────────────────────────────────────────────────
Unrecognized extensions dropped from resources:
 Condition                            
 └── http://example.com/blarg (1 time)
 Encounter                                                                      
 ├── http://cerner.com/long-url/extensions/estimated-financial-responsibility-am
 │   ount (1 time)                                                              
 └── http://example.com/blarg (2 times)                                         
────────────────────────────────────────────────────────────────────────────────
🚨 Resources skipped due to unrecognized modifier extensions: 🚨
 Encounter                          
 └── http://example.com/foo (1 time)
────────────────────────────────────────────────────────────────────────────────
⭐ All tasks completed successfully! ⭐
```

### Checklist

### DIFF
--- a/cumulus_etl/deid/scrubber.py
+++ b/cumulus_etl/deid/scrubber.py
@@ -4,11 +4,22 @@ import logging
 import tempfile
 from typing import Any
 
-from cumulus_etl import fhir
+import rich
+import rich.padding
+import rich.tree
+
+from cumulus_etl import common, fhir
 from cumulus_etl.deid import codebook, mstool, philter
+
+# Record of unknown extensions (resource type -> extension URL -> count)
+ExtensionCount = dict[str, dict[str, int]]
 
 
 class SkipResource(Exception):
+    pass
+
+
+class SkipValue(Exception):
     pass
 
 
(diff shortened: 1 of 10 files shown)


cumulus_etl/deid/scrubber.py
mikix, line 1: Sorry that there's some refactoring noise - I basically tweaked how the recursive scrubbing worked - basically just reduced the amount of full-object access each scrubber method had.

tests/data/mstool/output/Patient.1.json
mikix, line 1: These test files are for the ms-tool tests, which now no longer strips the extensions - so these changes just add the input extensions to the expected output.

dogversioning (approved):

PR #349, opened 2024-10-15
Add export URL & ETL version into completion table

mikix:
This should help track how we exported and processed the data.

### Checklist


tests/loaders/ndjson/test_bulk_export.py
```diff
@@ -755,6 +755,25 @@ async def test_retry_status_poll_then_success(self):
         )
 
 
+class TestBulkExportLogWriter(utils.AsyncTestCase):
```
mikix, line 758: Coverage tax

dogversioning (approved):

PR #351, opened 2024-10-21
Skip tasks or error out when resources are missing

mikix:
- If tasks are selected with --task or --task-filter and a needed resource is missing, we now error out.
- If tasks are not specifically selected, we now only run tasks for which resources are available (and error out if no resources exist at all)
- If --allow-missing-resources is passed, we skip both above checks and do historical behavior of running any given task with zero input rows (which pushes up a schema, cleans up the delta lake, and stamps the resource as complete in the completion tracking tables).

This addresses a lingering issue from #296, where we didn't adequately protect ourselves from incorrectly handling "empty set" inputs and wrongly marking them as complete. I've decided to go the simplest route and refuse to handle empty-sets unless specifically told to. I think this is appropriate given how unlikely empty sets are in the real world. (We see them with Cerner for some resources like ServiceRequest, but it's the exception not the rule.)

(An alternate solution that I did consider was looking at the bulk export log if it was present, and seeing what resources were exported - then continuing if the export log says it exported Patients, even if no patient files where present. But it seemed more complicated code-wise and more brittle if users are moving files around. The approach I went with requires a human in the loop to agree to allow empty sets. Or... someone just adds the --allow flag to their scripts and we lose this protection anyway - but that's at least not on me. :smile:)

### Checklist

### DIFF
--- a/cumulus_etl/loaders/fhir/ndjson_loader.py
+++ b/cumulus_etl/loaders/fhir/ndjson_loader.py
@@ -2,6 +2,8 @@
 
 import tempfile
 
+import cumulus_fhir_support
+
 from cumulus_etl import cli_utils, common, errors, fhir, store
 from cumulus_etl.loaders import base
 from cumulus_etl.loaders.fhir.bulk_export import BulkExporter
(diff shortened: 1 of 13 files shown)


cumulus_etl/loaders/fhir/ndjson_loader.py
```diff
@@ -37,7 +39,16 @@ def __init__(
         self.until = until
         self.resume = resume
 
-    async def load_all(self, resources: list[str]) -> base.LoaderResults:
+    async def detect_resources(self) -> set[str] | None:
+        if self.root.protocol in {"http", "https"}:
+            return None  # we haven't done the export yet!
```
dogversioning, line 44: nit - with this and the i2b2 tcp thing below, if i was coming into this codebase fresh, a little more context as to what this means might be helpful (i.e. it's we've got the URL of the place to get the data, not the data itself)

dogversioning (approved):

cumulus_etl/loaders/fhir/ndjson_loader.py
mikix: OK, I added some more context:

```
            # We haven't done the export yet, so there are no files to inspect yet.
            # Returning None means "dunno" (i.e. "just accept whatever you eventually get").
            return None
```

PR #352, opened 2024-10-23
feat: add new `init` subcommand

mikix:
This will create an empty table for each basic resource that we support. This is a convenience command to get up and running faster or to refresh your database schemas in case we change them in the future.

### Checklist

### DIFF
--- a/cumulus_etl/etl/init/cli.py
+++ b/cumulus_etl/etl/init/cli.py
@@ -1,0 +1,70 @@
+"""
+Initializes basic resource tables.
+
+Creates the tables if they don't exist and pushes up a basic schema.
+"""
+
+import argparse
+from collections.abc import Iterable
+
+from cumulus_etl import cli_utils, formats, store
+from cumulus_etl.etl import tasks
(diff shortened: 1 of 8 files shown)


cumulus_etl/etl/init/cli.py
```diff
@@ -0,0 +1,70 @@
...
+async def init_main(args: argparse.Namespace) -> None:
+    """Main logic for initialization"""
+    # record filesystem options like --s3-region before creating Roots
+    store.set_user_fs_options(vars(args))
+
+    output_root = store.Root(args.dir_output)
+
+    with cli_utils.make_progress_bar() as progress:
+        # Set up progress bar
+        total_steps = len(list(get_task_tables())) + 1  # extra one is format init
```
dogversioning, line 46: nit - can you reword this comment?

dogversioning (approved):

cumulus_etl/etl/init/cli.py
mikix: `extra 1 is initializing the formatter`

PR #354, opened 2024-10-28
Enable completion tracking by default

mikix:
- Removes the --write-completion opt-in flag, it is now always enabled.
- Requires export group name and timestamp information to be available, either from export log or from CLI.
- Updates some user docs, explaining how completion tracking expects to be fed data.
- Drops deprecated subcommand alias `chart-review` (old name for `upload-notes`)
- Bump pyarrow to 18.x (which allows compatibility with python 3.13)
- Bumps version to 2.0.0
- Tag docker releases with major number, so that someone who wants to be careful to avoid unexpected CLI breaks like this introduces could track their install that way.

This is the last big piece of #296

### Checklist


tests/data/simple/input/log.ndjson
mikix, line 1: This file added so that the "sample ETL runs" in our docs, which point at this folder, can continue working.

.github/workflows/docker-hub.yaml
mikix, line 1: This file's changes are not yet tested - let's hope it works? Else I can do some fix up PRs afterward. Not sure to best test this piece of the deployment mechanism (without affecting existing users of our docker images)

dogversioning: you could temporarily change the tagging strategy? then you could run it from this branch without disruption?

docs/bulk-exports.md
dogversioning, line 1: coherence check - do we need to add any updates anyplace else? should we link to the expectations from sections of the docs like sample runs to call out some of the new assumptions?

dogversioning (approved):

.github/workflows/docker-hub.yaml
mikix: I did this - and it did catch an issue, so good idea to push back against my yolo strategy :smile: - filed the docker stuff separately though in #355

docs/bulk-exports.md
mikix: OK yeah, I added a link to this bulk export doc from the sample run doc in the "more realistic command" section, like "and if you want real data, see this doc for how"

PR #357, opened 2024-11-04
ci: enable dependabot

mikix:
### Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,7 @@ dependencies = [
     "openai < 2",
     "oracledb < 3",
     "philter-lite < 1",
-    "pyarrow < 19",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -22,7 +22,7 @@ dependencies = [
     "openai < 2",
     "oracledb < 3",
     "philter-lite < 1",
-    "pyarrow < 19",
+    "pyarrow < 18",
```
mikix, line 25: I'm not 100% sure that dependabot will do something sane with the way our constraints are set up with maximums. Hopefully it will! But as a test, I'm also dropping pyarrow down a notch for the moment. Let's see what it does.

dogversioning (approved):

PR #363, opened 2024-11-18
fix: allow overwriting delta lake entries with same timestamp

mikix:
Without allowing this, we can't meaningfully update rows as ETL evolves to allow more content (_data and _url fields recently, maybe future allowed extensions, that sort of thing).

This does allow more data churn, but correctness takes priority. We could maybe get both if we inserted the ETL version into every row? But for now, this is an easy tweak.

### Checklist

### DIFF
--- a/cumulus_etl/formats/deltalake.py
+++ b/cumulus_etl/formats/deltalake.py
@@ -196,22 +196,30 @@ class DeltaLakeFormat(Format):
         if not has_last_updated_field:
             return None
 
-        # OK, the field exists (which is typical for any FHIR resource tables, as we provide a wide FHIR schema),
-        # so we want to conditionally update rows based on the timestamp.
+        # OK, the field exists (which is typical for any FHIR resource tables, as we provide a wide
+        # FHIR schema), so we want to conditionally update rows based on the timestamp.
         #
-        # We skip the update row if both the table and the update have a lastUpdated value and the update's value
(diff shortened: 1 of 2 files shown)


cumulus_etl/formats/deltalake.py
```diff
@@ -196,22 +196,30 @@ def _get_update_condition(schema: pyspark.sql.types.StructType) -> str | None:
...
-        # If we eventually decide that sub-second updates are a real concern, we can make it <= and
-        # additionally compare versionId. But I don't know how you extracted both versions so quickly. :)
+        # This uses less-than-or-equal instead of less-than when comparing the date, because
+        # sometimes the ETL will upload different content for the same resource as we update the
+        # ETL (for example, we allow-list yet another extension - we still want to re-upload the
+        # content with the new extension but same lastUpdated value). This does cause some needless
+        # churn on the delta lake side, but we'll have to live with that.
+        #
+        # If we eventually decide that sub-second updates are a real concern, we can additionally
+        # compare versionId. But I don't know how you extracted both versions so quickly. :)
```
dogversioning, line 213: yeah i don't think we're going to have this problem any time soon

dogversioning (approved):

cumulus_etl/formats/deltalake.py
mikix: But _correctness_ Maaaatt!

PR #364, opened 2024-11-19
feat(deid): drop valueString instead of passing it through

mikix:
Previously, we would let Observation.valueString (and its component form) through, optionally masking it with philter if you passed --philter in.

But it held PHI too often, and if we ever will use the field, it will likely be via ETL-side NLP. We don't need to keep the string intact into Athena.

So now, we unconditionally drop Observation.valueString, replacing it with a data-absent-reason extension marker (so consumers know it _was_ present).

### Checklist

### DIFF
--- a/cumulus_etl/deid/scrubber.py
+++ b/cumulus_etl/deid/scrubber.py
@@ -406,22 +406,30 @@ class Scrubber:
         if isinstance(value, str) and any(
             (
                 # Coding.display is clinically unnecessary but is useful for rendering.
-                # Since "code" is always present, downstream consumers can & should provide their own display label.
-                # But we don't remove it entirely, for cases where unexpected codes are used.
-                # Note that this will definitely over-scrub (like scrubbing "White" from the USCDI race extension),
-                # but again -- this display value is redundant and rather than try to be smart, we're safely dumb.
+                # Since "code" is always present, downstream consumers can & should provide their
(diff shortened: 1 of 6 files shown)


cumulus_etl/deid/scrubber.py
```diff
@@ -406,23 +406,31 @@ def _check_text(self, key: str, value: Any) -> Any:
...
 
+        if isinstance(value, str) and any(
+            (
+                # Observation.valueString has some clinical value, but often holds PHI.
+                # If we ever want to use this for clinical purposes, we should process it via NLP
+                # on the ETL side (like we do for DocumentReference attachments).
+                key == "valueString",
+            )
+        ):
+            raise MaskValue
```
dogversioning, line 432: i'm sure this is correct, but just for my own edification, where is this caught?

dogversioning (approved):

cumulus_etl/deid/scrubber.py
mikix: Fair question :smile: 

The loop of the scrubber is that it walks the JSON tree, and runs all these little scrubbing handlers that if they recognize something they should scrub, modify the value or raise an exception like this.

So the answer is: one layer up in the Scrubber code. See https://github.com/smart-on-fhir/cumulus-etl/blob/741fb34678425759df58b91e6ba4da433f74e509/cumulus_etl/deid/scrubber.py#L167

PR #369, opened 2024-12-26
docs: update jwks generation commands to separate public/private keys

mikix:
### Checklist

### DIFF
--- a/tests/loaders/ndjson/test_bulk_export.py
+++ b/tests/loaders/ndjson/test_bulk_export.py
@@ -3,6 +3,7 @@
 import contextlib
 import datetime
 import io
+import json
 import os
 import tempfile
 from unittest import mock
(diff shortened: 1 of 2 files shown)


tests/loaders/ndjson/test_bulk_export.py
mikix, line 1: Something changed in the json formatting library that something in our stack uses.

Previously, we were passing python objects to `respx` to return in mock http responses, then expecting a hardcoded number of bytes back - i.e. baking assumptions about the json encoding that would occur behind the scenes into our tests. Whoops.

So now the tests encode json to text themselves and keep track of the size.

mikix: This was unrelated to the docs change, I'm just using this PR as a convenient place to fix the tests.

PR #371, opened 2025-01-10
Switch --smart-jwks to --smart-key and accept PEM keys

mikix:
Some servers (notably Epic) support using a private key as an alternative to a JWK Set. So let's add support for that too.

It's a little fussier, but easy enough.

### Checklist


cumulus_etl/cli_utils.py
```diff
@@ -17,6 +17,7 @@ def add_auth(parser: argparse.ArgumentParser, *, use_fhir_url: bool = True):
     group = parser.add_argument_group("authentication")
     group.add_argument("--smart-client-id", metavar="ID", help="client ID for SMART authentication")
     group.add_argument("--smart-jwks", metavar="PATH", help="JWKS file for SMART authentication")
+    group.add_argument("--smart-pem", metavar="PATH", help="PEM file for SMART authentication")
```
mikix, line 20: Option B: deprecate `--smart-jwks` (but keep it around secretly) and replace it with `--smart-key`, and figure out whether it's a PEM or a JWKS in realtime -- maybe content sniffing, maybe as lazy as file suffixes, but I don't think PEM suffixes are fully normalized.

I went with the extremely explicit route, but maybe that's worse UX - I'm not sure.

cumulus_etl/fhir/fhir_auth.py
```diff
@@ -128,6 +131,42 @@ async def _get_token_endpoint(self, session: httpx.AsyncClient) -> str:
...
+                # supports it. Since 384 is practically as secure as 512, but more common, we'll
+                # use that.
+                return "RS384", jwk_key
+            elif jwk_key.get("kty") == "EC" and jwk_key.get("crv") == "P-256":
+                return "ES256", jwk_key
+            elif jwk_key.get("kty") == "EC" and jwk_key.get("crv") == "P-384":
+                return "ES384", jwk_key
+            elif jwk_key.get("kty") == "EC" and jwk_key.get("crv") == "P-521":
+                # Yes, P-521 is not a typo, it's what the curve is called.
+                # The curve uses 521 bits, but it's hashed with SHA512, so it's called ES512.
```
dogversioning, line 154: case study in 'im glad i don't work in cryptography' right there

cumulus_etl/cli_utils.py
dogversioning: I do like hiding the complexity of 'which key format am i using' from the CLI - up to you if you wanna try it or not.

dogversioning (approved):

cumulus_etl/cli_utils.py
mikix: I can work on that :+1:

mikix: Done!

PR #374, opened 2025-01-16
feat: add inlining options and a new `inline` command

mikix:
Inlining here means downloading attachment URLs and insert them as attachment data, allowing you to download the attachment just once, at the cost of extra storage space.

This is often useful in preparation for NLP tasks, where you want to often refer back to DocumentReference clinical notes, but do not want to constantly deal with the flakiness of an EHR server. (Or losing access to that server.)

New features:
- There is a new `inline` command that can inline an existing folder of NDJSON.
- The `export` and `etl` commands both now accept an opt-in flag to inline data after performing a bulk export.
- If the server provides a Retry-After header that is a timestamp (rather than a number of seconds), we now parse that correctly.

Behavior changes:
- If the server gives us a 429 error, we now log it as an error message, and don't log a successful progress call.
- If the server gives us a 429 error, we use the next exponential backoff delay instead of hard coding 60 seconds as the delay.
- If the server gives us a Retry-After header on an error message, we no longer unconditionally accept and use it. Rather the requested delay is capped by our next exponential backoff delay. That is, the server's Retry-After time will be used if it's LESS than our next backoff, but if it's longer, we'll still use our own backoff. This is lightly hostile, but (a) it's only done on error cases, (b) our backoff times are generous, and counted in minutes not seconds, and (c) it lets us guarantee a max delay time for callers.
- Instead of retrying on 429 and ALL 5xx errors, there's a specific list of error codes that we retry on. Currently it's 408, 429, 502, 503, and 504.
- Have the bulk exporter time out after 30 days, rather than one. We've seen Epic exports take a couple weeks.

### Checklist

### DIFF
--- a/cumulus_etl/fhir/fhir_client.py
+++ b/cumulus_etl/fhir/fhir_client.py
@@ -1,8 +1,10 @@
 """HTTP client that talk to a FHIR server"""
 
 import argparse
+import asyncio
+import email
 import enum
-from collections.abc import Iterable
+from collections.abc import Callable, Iterable
 from json import JSONDecodeError
 
 import httpx
@@ -27,6 +29,9 @@ class FhirClient:
 
     See https://hl7.org/fhir/smart-app-launch/backend-services.html for details.
     """
+
+    # Limit the number of connections open at once, because EHRs tend to be very busy.
+    MAX_CONNECTIONS = 5
 
     def __init__(
         self,
@@ -71,8 +76,7 @@ class FhirClient:
         self._capabilities: dict = {}
 
     async def __aenter__(self):
-        # Limit the number of connections open at once, because EHRs tend to be very busy.
-        limits = httpx.Limits(max_connections=5)
+        limits = httpx.Limits(max_connections=self.MAX_CONNECTIONS)
         timeout = 300  # five minutes to be generous
         # Follow redirects by default -- some EHRs definitely use them for bulk download files,
         # and might use them in other cases, who knows.
@@ -86,7 +90,15 @@ class FhirClient:
             await self._session.aclose()
 
     async def request(
-        self, method: str, path: str, headers: dict | None = None, stream: bool = False
+        self,
+        method: str,
+        path: str,
+        headers: dict | None = None,
+        stream: bool = False,
+        retry_delays: Iterable[int] | None = None,
+        request_callback: Callable[[], None] = None,
+        error_callback: Callable[[errors.NetworkError], None] = None,
+        retry_callback: Callable[[httpx.Response, int], None] = None,
     ) -> httpx.Response:
         """
         Issues an HTTP request.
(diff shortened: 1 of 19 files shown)


cumulus_etl/fhir/fhir_client.py
```diff
@@ -86,7 +90,15 @@ async def __aexit__(self, exc_type, exc_value, traceback):
             await self._session.aclose()
 
     async def request(
```
mikix, line 92: Let me lightly explain some of the refactoring here.

Before, FhirClient.request was a single request. The retry logic was in the bulk export code, which retried on both error cases and the expected `202` status code for "still working on it".

Now, I've separated the bulk export retry logic into error-path and 202-path, and moved the error-path into this FhirClient.request method. So now any code making requests to the server can enable retries if they pass in a `retry_delays` array of retry timeouts.

.github/workflows/ci.yaml
```diff
@@ -113,7 +113,7 @@ jobs:
           # Compare results
           export OUTDIR=$DATADIR/run-output/covid_symptom__nlp_results
           sudo chown -R $(id -u) $OUTDIR
-          sed -i 's/"generated_on": "[^"]*", //g' $OUTDIR/*.ndjson
+          sed -i 's/"generated_on":"[^"]*",//g' $OUTDIR/*.ndjson
```
mikix, line 116: Adjusted to remove expected whitespace because this PR also switches to a more condensed NDJSON output mode for the ETL that removes whitespace (in an attempt to reduce bloat a little bit when inlining).

pyproject.toml
```diff
@@ -21,7 +21,7 @@ dependencies = [
     "openai < 2",
     "oracledb < 3",
     "philter-lite < 1",
-    "pyarrow < 19",
+    "pyarrow < 20",
```
mikix, line 24: Unrelated, I just stole this from a [dependabot PR](https://github.com/smart-on-fhir/cumulus-etl/pull/375) that was consistently failing CI for no good reason.

cumulus_etl/inliner/writer.py
```diff
@@ -0,0 +1,46 @@
...
+
+class OrderedNdjsonWriter:
+    """
+    Convenience context manager to write multiple objects to a ndjson file in order.
+
+    Specifically, it will keep the output in the intended order, even if lines are provided
+    out of order.
+
+    Note that this is not atomic - partial writes will make it to the target file.
+    And queued writes may not make it to the target file at all, if interrupted.
```
dogversioning, line 12: this makes me nervous - i think it's probably right to push out this complexity til later, but this feels like something that might cause problems later

dogversioning (approved):
Mostly skimmed the tests, otherwise looks good

cumulus_etl/inliner/writer.py
mikix: Fair point. In this context, I think it's safe. We write the inlined files to a temp location, then [atomically swap](https://github.com/smart-on-fhir/cumulus-etl/blob/7f2d0ab0136ed6d63691378fe82e3eb05ed5d134/cumulus_etl/inliner/inliner.py#L148) out the old one when done, assuming there were no errors.

cumulus_etl/inliner/inliner.py
```diff
@@ -0,0 +1,217 @@
...
+                    mimetypes=mimetypes,
+                    output=output,
+                    stats=stats,
+                    progress=progress,
+                    progress_task=progress_task,
+                ),
+                # Look at twice the allowed connections - downloads will block, but that's OK.
+                # This will allow us to download some attachments while other workers are sleeping
+                # because they are waiting to retry due to an HTTP error.
+                peek_at=fhir.FhirClient.MAX_CONNECTIONS * 2,
```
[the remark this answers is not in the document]
mikix: > should this be configurable?

The times-two approach feels too nitty gritty to configure. It's just a way to keep some balls spinning while others are blocked on the timeouts. I feel like this number would scale with the length of the timeouts, not server performance. So if we wanted to configure this, you'd probably also want to configure the timeouts. I dunno, too low level in my head.

cumulus_etl/fhir/fhir_client.py
```diff
@@ -28,6 +30,9 @@ class FhirClient:
     See https://hl7.org/fhir/smart-app-launch/backend-services.html for details.
     """
 
+    # Limit the number of connections open at once, because EHRs tend to be very busy.
+    MAX_CONNECTIONS = 5
```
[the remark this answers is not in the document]
mikix: Possibly? This was set by me way back when after doing some small testing on our Cerner staging instance. I didn't really re-examine this default value this time around (i.e. didn't test different numbers during inlining). This current value is not scientific as is, and definitely could change based on the server size / load.

But also... I dunno, this inliner is the first place it'll super matter (other network request paths aren't fully parallel yet). And it's such a long job, that tweaking a config option like this to improve performance by 10% or something feels like a lot to ask for a casual user.

But I'm not opposed to allowing customization. The above is just an explanation of why I haven't done it yet. I might want to wait until we notice a problem with the default, or a problem with the speed of inlining jobs that we think we can push, just to delay adding complexity.

When we get access to BCH Epic, I'd like to do some speed tests there with different numbers too.

cumulus_etl/inliner/inliner.py
```diff
@@ -0,0 +1,217 @@
...
+import fsspec
+import rich.progress
+import rich.table
+
+from cumulus_etl import cli_utils, common, errors, fhir, store
+from cumulus_etl.inliner import reader, writer
+
+
+@dataclasses.dataclass
+class Stats:
```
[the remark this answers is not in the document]
mikix: Sure :+1:

PR #377, opened 2025-02-05
fix: a couple extensions were incorrectly being warned about

mikix:
Due to a missing comma, two known-but-we-want-to-silently-ignore-them extensions were being warned about instead of silently ignored:
- http://hl7.org/fhir/StructureDefinition/patient-mothersMaidenName
- http://hl7.org/fhir/StructureDefinition/rendered-value

### Checklist

### DIFF
--- a/cumulus_etl/deid/scrubber.py
+++ b/cumulus_etl/deid/scrubber.py
@@ -328,7 +328,7 @@ class Scrubber:
                 "http://hl7.org/fhir/StructureDefinition/geolocation",
                 "http://hl7.org/fhir/StructureDefinition/iso21090-TEL-address",
                 "http://hl7.org/fhir/StructureDefinition/patient-birthPlace",
-                "http://hl7.org/fhir/StructureDefinition/patient-mothersMaidenName"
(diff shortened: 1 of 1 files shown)


cumulus_etl/deid/scrubber.py
```diff
@@ -328,7 +328,7 @@ def _check_extensions(
                 "http://hl7.org/fhir/StructureDefinition/geolocation",
                 "http://hl7.org/fhir/StructureDefinition/iso21090-TEL-address",
                 "http://hl7.org/fhir/StructureDefinition/patient-birthPlace",
-                "http://hl7.org/fhir/StructureDefinition/patient-mothersMaidenName"
+                "http://hl7.org/fhir/StructureDefinition/patient-mothersMaidenName",
```
mikix, line 331: I looked at ruff config options for this - and they (the ISC ones) conflict with the formatter. It seems that implicit concatenation across lines like this is how the style wants such lines.

So it's not easy to safeguard against this in the future via ruff check... Just gotta be careful I guess.

PR #379, opened 2025-02-14
feat(upload-notes): add --highlights option to directly annotate

mikix:
The given words will be pre-annotated, much like with a cTAKES dictionary, but without having to bother with cTAKES or bsv files and thus whether the terms are clinical with a CUI etc. You can just search for "Mike" if you want.

### Checklist

### DIFF
--- a/cumulus_etl/upload_notes/cli.py
+++ b/cumulus_etl/upload_notes/cli.py
@@ -3,6 +3,7 @@
 import argparse
 import asyncio
 import datetime
+import re
 import sys
 from collections.abc import Collection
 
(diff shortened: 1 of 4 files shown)


tests/upload_notes/test_upload_labelstudio.py
```diff
@@ -85,7 +87,6 @@ def test_basic_push(self):
                             # as it was not in our initial CUI mapping (in push_tasks)
                             {
                                 "from_name": "mylabel",
-                                "id": "ctakes0",
```
mikix, line 88: I realized that `id` isn't a real field in this part, afaict - I don't know why it was added in the first place. But removing it lets me not keep track of a counter or what type of object is being serialized, so it's gone baby.

cumulus_etl/upload_notes/cli.py
```diff
@@ -272,7 +297,7 @@ def define_upload_notes_parser(parser: argparse.ArgumentParser) -> None:
         "--philter",
         choices=[PHILTER_DISABLE, PHILTER_REDACT, PHILTER_LABEL],
         default=PHILTER_REDACT,
-        help="Whether to use philter to redact/tag PHI",
+        help="Whether to use philter to redact/tag PHI (default is redact)",
```
mikix, line 300: Unrelated, just wanted to clarify while I was here.

dogversioning (approved):

PR #381, opened 2025-02-18
fix(upload-notes): if we are pushing a lot of LS notes, batch them

mikix:
This avoids timeout errors if we're trying to send a thousand or so notes.

As part of this, I used our internal batching method and upgraded it a bit:
- Can now accept synchronous iteraters
- There is now a version of it that does not require consumers to unpack a tuple (i.e. a version that does not support parallel streams of data which the normal ETL task version does need)
- Updated its docstrings

### Checklist


cumulus_etl/batching.py
```diff
@@ -0,0 +1,118 @@
+"""Iteration utilities to support batching values."""
```
mikix, line 1: This is mostly the same as the deleted `batching.py` file, git just figured they were so different it didn't map the move.

Changes:
- Renamed `batch_iterate` to `batch_iterate_streams` and expand its docstring
- Added a `batch_iterate` that returned an iterator of actual batches (instead of tuples of batches, which is what `batch_iterate_streams` does - for most normal use cases, you don't need streaming support - that's a complexity that only the ETL tasks need)
- Added `_force_aiter` which upgrades a boring old sync iterable into an async iterable. So now these batching methods can take either.

PR #382, opened 2025-02-25
WIP docs: refactor the initial setup documentation to be simpler

mikix:
The doc refactor mostly does the following, in addition to some cleanup:
- Separates out the "local test run" from the "production AWS" setup
- Emphasizes the `init` step
- Avoids encouraging a full git clone, but instead just grabbing the compose.yaml file, since that's all we need.

Additionally, there are some Docker behavior changes:
- Adds a check to the cTAKES container setup to ensure a UMLS key is defined, and then the docs only bother mentioning the UMLS key in the NLP pathways, rather than all pathways (and the Docker warning about not defining a UMLS key even when simply running base ETL goes away)
- Sets Docker to always pull the latest ETL image.

### Checklist


.github/workflows/ci.yaml
```diff
@@ -92,7 +92,7 @@ jobs:
       - name: Download NLP images
         run: |
           # Just start the specific services we need, to conserve CI disk space
-          docker compose up -d --quiet-pull cnlpt-negation ctakes-covid
+          docker compose up --wait --quiet-pull cnlpt-negation ctakes-covid
```
mikix, line 95: Using `--wait` makes a healthcheck failure much more visible to the user - and now that we added a cTAKES healthcheck, it makes sense to encourage this flow.

compose.yaml
```diff
@@ -26,6 +26,7 @@ services:
   cumulus-etl-base:
     extends: common-base
     image: smartonfhir/cumulus-etl:latest
+    pull_policy: always  # ensure folks are using the lastest release
```
mikix, line 29: This is a big change I wanted to flag. It's not necessary to the rest of this PR, but I made it while I was thinking about the install / deploy flow.

This means that every time a user runs `docker compose run cumulus-etl`, Docker will pull down `:latest` first.

Pros:
- users always get the latest, which has been an issue in the past

Cons:
- risk of bugs sneaking out to everyone all of a sudden (I think we've got a good track history with bugs, and we have lots of CI, but it's still a real risk -- flip side of this is that it's easy to deploy any fixes)
- custom builds of ETL with patches may get overridden unless the patcher also goes in and comments out this line (this is not necessarily a supported flow, but is something some folks have been forced to do as stop gaps)

Is this worth it? What's your vibe?

While we're considering the deployment flow, I'd also like to mention here that further down, I swap out the instructions to `git clone` the repo in order to get the `compose.yaml` with instructions to just directly `wget` the compose.yaml` file from github.

I did this because I felt like it was confusing to grab the whole repo for just one file and then that Docker file ignores the repo and goes to Docker Hub to actually grab the code run. It felt like one too many copies of the ETL sitting around.

But making that change made me think about "how do we encourage users noticing updates to compose.yaml?"
- I'm not entirely convinced that a git clone helps there. The person would have to know (a) that `git pull` will get new stuff (likely for the historical kind of devs that do this, but still, the instructions are simple enough for non-devs to try it) and (b) more likely, the person would not think to pull new stuff.
- I had a thought that if we ever updated the compose.yaml in a way that we wanted to ensure folks got, we could add a variable in the file that sets something like `CUMULUS_COMPOSE_VERSION=1` and then add a check in the ETL code that would warn (or error out if severe enough) about updating the compose file and how. (and treating the absence of that variable as version zero) This idea works especially well if we keep the `always` pull policy.

compose.yaml
```diff
@@ -69,7 +70,11 @@ services:
     image: smartonfhir/ctakes-covid:1.1.1
     environment:
       - ctakes_umlsuser=umls_api_key
-      - ctakes_umlspw=$UMLS_API_KEY
+      - ctakes_umlspw=${UMLS_API_KEY:-}
+    healthcheck:
+      test: echo 'Must set the UMLS_API_KEY environment variable.' && test -n "${UMLS_API_KEY:-}"
```
mikix, line 75: This message only shows up if you manually go and do a `docker inspect` :shrug: But it's something.

compose.yaml
dogversioning: how would you feel about pulling a non-latest tag, and having that be decoupled from latest? that would give some flexibility. I don't hate the pure CI path, and I feel like the ETL is pretty mature, but it reduces the 'oh no i have to cut a release right now' pressure.

On the pulling stuff: Would you consider having a different repo with just the compose file/some light documentation? that would enable the git clone as a minimal thing. But i don't hate the 'just clone the repo' approach - i don't feel like that's terribly uncommon, and is useful if for some reason someone did want to modify it (and run it in a dev-centric way)

docs/index.md
```diff
@@ -11,7 +11,7 @@ has_children: true
 
 ETL stands for "extract, transform, load."
 
-1. Cumulus ETL first **extracts** data from the hospital servers
+1. Cumulus ETL first **extracts** data from the EHR servers
```
dogversioning, line 14: nit: with this edit, i think `data from EHR servers` is more natural

dogversioning (approved):

compose.yaml
mikix: Crazy brainstorm: the ETL could bundle in the compose.yaml file. Then we have a subcommand like `docker run cumulus-etl spit-out-compose-file` that drops it in your current folder....  Is that crazy?

dogversioning: `it's awkward to use the git folder as your working directory. ` - maybe we should natter about this? i don't quite get it.

mikix: OK the current commit does two new things:
- Drops tagging with `:latest` and `:<version>` but keeps `:main` - we can do the stable tags manually for now and/or set up some CI triggers
- Edits the "Production Setup" instructions to use `git` and a throwaway line about running `git pull` for updates - I kept the "Local Test Setup" instructions using `wget` since there, we don't care about updates to the file and I wanted to keep an uncluttered working folder. Eh?

dogversioning: yep, i buy it

PR #384, opened 2025-03-05
fix(upload-notes): improve --highlight matching

mikix:
- No longer ignore matches in previous docs when grouping (whoops)
- Instead of labeling a found term with the term itself, just mark it with a generic "Keyword" label to be less noisy in the Label Studio interface (Andy request)
- Escape regex characters in highlight terms
- Improve boundary matching when a term includes non-word characters (like a + or -)

### Checklist

### DIFF
--- a/cumulus_etl/upload_notes/cli.py
+++ b/cumulus_etl/upload_notes/cli.py
@@ -162,13 +162,20 @@ def add_highlights(notes: Collection[LabelStudioNote], args: argparse.Namespace)
 
     common.print_header("Highlighting notes...")
 
-    re_terms = [re.compile(rf"\b{term}\b", re.IGNORECASE) for term in highlights]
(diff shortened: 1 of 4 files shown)


cumulus_etl/upload_notes/cli.py
```diff
@@ -240,11 +247,9 @@ def group_notes_by_encounter(notes: Collection[LabelStudioNote]) -> list[LabelSt
                 match.end += offset
                 grouped_ctakes_matches.append(match)
 
-            for label, spans in note.highlights.items():
-                grouped_highlights[label] = []
```
mikix, line 244: This line here was a nasty bug I missed - it would remove highlighted terms from previous docs in the group for this label.

dogversioning (approved):

PR #385, opened 2025-03-06
build: add script to update stable-release Docker tags

mikix:
Just run ./tag-release.sh and it's updated!

### Checklist

### DIFF
--- a/tests/inliner/test_inliner.py
+++ b/tests/inliner/test_inliner.py
@@ -107,7 +107,7 @@ class TestInliner(TestInlinerBase):
 
         await self.run_inline()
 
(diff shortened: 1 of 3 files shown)


tests/inliner/test_inliner.py
```diff
@@ -107,7 +107,7 @@ async def test_happy_path(self):
 
         await self.run_inline()
 
-        self.assertEqual(["docrefs.jsonl", "reports.jsonl"], os.listdir(self.input_dir))
+        self.assertEqual({"docrefs.jsonl", "reports.jsonl"}, set(os.listdir(self.input_dir)))
```
mikix, line 110: I noticed a build fail due to the order of the results here being different.

dogversioning (approved):
love it

PR #393, opened 2025-03-20
feat: write bulk export log v2 format (with manifest events)

mikix:
We don't ever request partial manifest pages, so this is just a logging format change to emit a few new events.

### Checklist

### DIFF
--- a/cumulus_etl/loaders/fhir/export_log.py
+++ b/cumulus_etl/loaders/fhir/export_log.py
@@ -237,13 +237,37 @@ class BulkExportLogWriter:
 
     def status_complete(self, response: httpx.Response):
         response_json = response.json()
+        transaction_time = response_json.get("transactionTime")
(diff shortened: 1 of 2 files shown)


cumulus_etl/loaders/fhir/export_log.py
```diff
@@ -237,13 +237,37 @@ def status_progress(self, response: httpx.Response):
 
     def status_complete(self, response: httpx.Response):
```
mikix, line 238: These changes are informed by the wiki page for this log format: https://github.com/smart-on-fhir/bulk-data-client/wiki/Bulk-Data-Export-Log-Items/

And specifically, the recent changes to support a new feature of bulk exporting: partial manifests (which let you get some export results before the whole export has finished). The ETL client doesn't support this yet, so this logging support is bare bones. But it adds the newly expected events to the log.

dogversioning (approved):

PR #395, opened 2025-03-25
build: push to pypi upon release

mikix:
Also, only push docker images upon release. And drop upper limits on dependencies, to make pip users' lives easier.

### Checklist

### DIFF
--- a/.github/workflows/docker-hub.yaml
+++ b/.github/workflows/docker-hub.yaml
@@ -1,12 +1,8 @@
 name: Docker Hub
 
-# Currently, we publish every commit to main into Docker hub.
-# And then manually tag stable releases.
-# In the future, we may want to trigger off of actual release tags.
 on:
-  push:
-    branches:
-      - main
+  release:
+    types: [created]
 
 jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/docker-hub.yaml
```diff
@@ -25,10 +21,10 @@ jobs:
         id: meta
         uses: docker/metadata-action@v5
         with:
-          flavor: latest=false
+          flavor: latest=true
           images: smartonfhir/cumulus-etl
           tags: |
-            type=ref,event=branch
+            type=pep440,pattern={{major}},value=${{ env.ETL_VERSION }}
```
mikix, line 27: This drops the docker build for every single commit (the `:main` tag). I could have kept it, but it felt less useful now that we have real releases... Do you like the idea of keeping them?

dogversioning (approved):

.github/workflows/docker-hub.yaml
dogversioning: Yeah, if you're commiting to versions, the docker versions should match the pypi version.

PR #398, opened 2025-04-02
fix: report an accurate bulk export duration in the logs

mikix:
### Checklist

### DIFF
--- a/cumulus_etl/loaders/fhir/export_log.py
+++ b/cumulus_etl/loaders/fhir/export_log.py
@@ -321,7 +321,7 @@ class BulkExportLogWriter:
 
     def export_complete(self):
         timestamp = common.datetime_now(local=True)
-        duration = (timestamp - self._start_time) if self._start_time else 0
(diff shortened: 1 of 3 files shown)


cumulus_etl/loaders/fhir/export_log.py
```diff
@@ -321,15 +321,15 @@ def download_error(self, file_url: str | httpx.URL, exc: Exception):
...
+        duration = (timestamp - self._start_time) if self._start_time else None
         self._event(
             "export_complete",
             {
                 "files": self._num_files,
                 "resources": self._num_resources,
                 "bytes": self._num_bytes,
                 "attachments": None,
-                "duration": duration.microseconds // 1000,
+                "duration": duration // datetime.timedelta(milliseconds=1) if duration else 0,
```
mikix, line 332: Boneheaded mistake :roll_eyes: 

`.microseconds` is just the microseconds field, not the total time of the duration in microseconds.

dogversioning (approved):

PR #408, opened 2025-04-24
build: set version dynamically from tag rather than __init__.py

mikix:
This makes it easier to release versions without needing to update the hardcoded version in the repo. Just release from github when it makes sense to.

### Checklist


dogversioning (approved):
If this works, i'd love to copy this to the other big repos

PR #410, opened 2025-04-28
feat(upload-notes): support DiagnosticReport resources as well

mikix:
- Add a lot more column headers we look for in ID files (not just docref_id, but document_ref, documentreference_id, diagnosticreport_ref, etc)
- Support columns that have a mix of DxReports and DocRefs
- When downloading from a FHIR server, also download DiagnosticReport resources

### Checklist


cumulus_etl/fhir/fhir_utils.py
```diff
@@ -244,36 +244,46 @@ async def _get_docref_note_from_attachment(client: "FhirClient", attachment: dic
...
-    note = _get_cached_docref_note(docref)
+    note = _get_cached_note(resource)
     if note is not None:
         return note
 
-    attachments = [content["attachment"] for content in docref["content"]]
+    match resource["resourceType"]:
+        case "DiagnosticReport":
+            attachments = resource.get("presentedForm", [])
+        case "DocumentReference":
```
dogversioning, line 279: How would you feel about moving the expected note containing types to an enum or a set or something like that?

dogversioning (approved):

cumulus_etl/fhir/fhir_utils.py
mikix: I'm not opposed to that, but "what you do with that information next" is often unique to the task the code is doing. Like here, in the highlighted code, how would a set variable like `NOTE_TYPES = {"DiagnosticReport", "DocumentReference"}` affect the `match` tree?

mikix: Unless maybe there's a like a `notes_utils.py` class that has a swiss army knife of utilities - `get_attachments_from_note` or something like that?

mikix: But actually, I guess that is what the methods in this file sort of are becoming.

mikix: Well an enum (or set of constant vars) might make sense for resource type names in general, to avoid typos. Hopefully tests catch typos, but why not defense in depth.

I'll have that in the back of my mind for the future / a more explicit set of note-oriented utils. But I am emotionally ready to just land the current set of fixes and not tweak them.

I already have to fix why the tests are failing in CI but not locally :frowning:

PR #411, opened 2025-04-28
feat(etl): support gzip input files (*.ndjson.gz)

mikix:
Now that cumulus-fhir-support can return *.gz files, we need a little extra support in the ETL to process them, because the MS tool doesn't know how to read them.

So we have to decompress them before pushing data through our pipeline.

### Checklist

### DIFF
--- a/cumulus_etl/loaders/fhir/ndjson_loader.py
+++ b/cumulus_etl/loaders/fhir/ndjson_loader.py
@@ -1,5 +1,8 @@
 """Ndjson FHIR loader"""
 
+import gzip
+import os
+import shutil
 import tempfile
 
 import cumulus_fhir_support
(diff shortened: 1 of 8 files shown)


cumulus_etl/loaders/fhir/ndjson_loader.py
```diff
@@ -85,9 +88,21 @@ async def load_resources(self, resources: set[str]) -> base.LoaderResults:
...
 
+    @staticmethod
+    def _decompress_file(path: str):
+        if not path.casefold().endswith(".gz"):
+            return
+        target_path = path[:-3]
+        with gzip.open(path, "rb") as f_in:
+            with open(target_path, "wb") as f_out:
+                shutil.copyfileobj(f_in, f_out)
+        os.remove(path)
```
dogversioning, line 104: This is probably a worse structural overhead thing, but - rather than replacing a compressed file with the unzipped file, for disk space usage a pattern like 'for each file in files: unzip it, do all the processing, and then remove the unzipped one before continuing' might be nice?

mikix: Ah fair, but this code flow is only for "copy the original file (maybe zipped) to a tmpdir, then process it" - so we already are working in a separate tmpdir and not the original input files. So no need to keep the original zip file in that case - we won't use it and it's not the user's actual copy of the file.

mikix: I was a little shocked that there isn't like a `shutil` utility method to just decompress a file as a one-liner. I guess questions like this are why not, but :shrug: - I'm finding that the "how to deal with compressed files" part of Python is all very low level with almost no abstraction layers.

PR #414, opened 2025-05-12
feat(upload-notes): disable Philter by default

mikix:
It might surprise folks to see notes affected like this by default. At BCH, we typically disable it all the time anyway. Let's go with the simple path by default, but allow it to be enabled.

### Checklist


dogversioning (changes requested):
this is fine from a technical perspective but i think we need to have a conversation internally from a legal perspective if we can ship something where the de-id mode is off by default before we put it in. if that's already happened i'll rescind my block.

mikix:
> this is fine from a technical perspective but i think we need to have a conversation internally from a legal perspective if we can ship something where the de-id mode is off by default before we put it in. if that's already happened i'll rescind my block.

I'm more interested in turning off cTAKES by default, so I don't need to die on the `philter` hill. Let me post in slack about this.

PR #415, opened 2025-05-14
WIP

mikix:
### Checklist


.github/workflows/pypi.yaml
```diff
@@ -23,7 +23,8 @@ jobs:
     - name: Set version from tag
       run: |
         ETL_VERSION=$(echo $GITHUB_REF_NAME | sed 's/^v//')
-        sed -i "s/0\.0\.0/$ETL_VERSION/" cumulus_etl/__init__.py
+        sed -i "s/1\!0\.0\.0/$ETL_VERSION/w changes.txt" cumulus_etl/__init__.py
+        [ -s changes.txt ] || exit 1  # validate that we successfully set the version
```
mikix, line 27: While I was here, I also tweaked the release version code to match what I did in data-metrics:
- Use an epoch for the development/git version, so that it will always be higher than the releases
- Confirm that we did correctly regex in the release pipeline, to try and avoid a future mixup

cumulus_etl/http.py
```diff
@@ -0,0 +1,213 @@
+"""HTTP helper methods"""
```
mikix, line 1: This file is mostly moved code. Here are all the changes I believe:
- tweaked how authentication is caught - brought it out of the old equivalent of `_request_once` and put it into the retry loop - it felt more honest to only actually make one request in `_request_once`.
- Look for a few more possible error messages inside an error reponse payload (this lets us simplify some of the oauth2 error handling code and just have it be done here)
- Catch httpx transport errors (like DNS issues) and tranform them into a errors.FatalNetworkError exception, so that the rest of the code only needs to catch one kind of error.

dogversioning (approved):

PR #419, opened 2025-06-03
feat(etl): read local Medication resources when processing MedReqs

mikix:
This lets the user pre-download Medication resources and present them to the ETL alongside the MedicationRequest resources. Thus avoiding a download step and letting the ETL be disconnected from the FHIR server during the ETL.

### Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ dependencies = [
     "aiobotocore < 2.22.0",  # FIXME: temp hotfix for dependency version madness - remove later
     "ctakesclient >= 5.1",
     "cumulus-fhir-support >= 1.2",
-    "delta-spark >= 3.2.1",
+    "delta-spark >= 3.2.1, < 4",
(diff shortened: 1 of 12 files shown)


pyproject.toml
```diff
@@ -5,7 +5,7 @@ dependencies = [
     "aiobotocore < 2.22.0",  # FIXME: temp hotfix for dependency version madness - remove later
     "ctakesclient >= 5.1",
     "cumulus-fhir-support >= 1.2",
-    "delta-spark >= 3.2.1",
+    "delta-spark >= 3.2.1, < 4",
```
mikix, line 8: Unrelated to main PR - I just noted that delta-spark is working on a v4 and it will require some changes in our code (mostly around which hadoop-aws module to pull in -- we'll have to change it to a new version number). For delta-spark, it's probably wise to keep gating the major version, even though that's annoying for users in the pip context.

PR #422, opened 2025-06-30
feat(upload-notes): add --grouping={encounter,none} flag

mikix:
This lets you turn off grouping notes by encounter, if that is not helpful for the kind of chart review you're doing.

Note that we keep using the "enc_id" field in Label Studio as the unique task ID field, to avoid breaking uploading to existing projects. We just push a ref like DocumentReference/A to that field in the case we aren't grouping by encounter. (just like we do when grouping by encounter but there isn't an encounter!)

### Checklist


cumulus_etl/upload_notes/labelstudio.py
```diff
@@ -122,12 +123,14 @@ def _format_task_for_note(self, note: LabelStudioNote) -> dict:
         task = {
             "data": {
                 "text": note.text,
+                # This "enc_id" name is historical - a better name is "grouping_id" or "unique_id".
+                # We use this enc_id field to determine which notes to override during upload.
+                # It's based off of whatever note grouping method was used when uploading.
+                "enc_id": note.unique_id,
```
mikix, line 129: This is the weirdest part of this PR - my thinking is that it's more important to have this tool keep working well for existing projects than for the metadata field on the Label Studio to be ideal. But I could be convinced off that.

dogversioning: personally i'm in favor of breaking this, or updating to a good id field and having a way to select an old one instead via a cli arg if you really want backwards compatibility.

dogversioning (approved):

PR #427, opened 2025-07-10
fix: error out if we detect a new PHI folder for same output folder

mikix:
This adds a new field to the codebook (`codebook_id`) that gets set on codebook load (and backfilled if not present). This lets us identify a PHI folder/codebook in a safe way.

Then each JobConfig writes that codebook ID out, and we confirm the most recent JobConfig and the current PHI folder match.

I did it this way (vs just comparing PHI dirs) because I wanted to be a little future-proof against someone moving the PHI dir, or mounting it in a different way if it's local, or whatever.

### Checklist


cumulus_etl/etl/config.py
```diff
@@ -121,3 +125,38 @@ def as_json(self):
...
+    if not saved_codebook_id:
+        return
+
+    # And compare against the new PHI dir
+    if saved_codebook_id != codebook_id:
+        errors.fatal(
+            f"The output folder '{output_root.path}' is already connected "
+            "to a different PHI folder. "
+            "You must always use the same PHI folder for a given output folder.",
+            errors.WRONG_PHI_FOLDER,
```
dogversioning, line 161: nit, ignorable: Putting myself in a user's shoes for a second, if I got this error message, I'm not sure I would 100% know what to do about it.

I don't know what I would do about it without modifying the data model - like, if you had the last path of the PHI, you could say 'it was used for PHI in location (y), though that may have changed if it was moved.' I'm not sure if saying something like 'look for a log with the string xxx' might also be helpful?

dogversioning (approved):

cumulus_etl/etl/config.py
mikix: Ah totally fair - we do actually already record the phi folder in the same config file. So I've added a bit of logic to pull that out and print it too. :+1:

PR #431, opened 2025-08-11
feat: add initial irae NLP task

mikix:
This is very alpha quality so far, the prompt and response formats are not settled.

But, adding initial support here lets us refactor a few things ahead of time:
- Drop hftest task - it was just a placeholder for a "real" task like irae, so now that we have one, let's drop this.
- Add some common "talk to an OpenAI API" task infrastructure, now used by irae and the covid tasks that talk to gpt.
- When sending to OpenAI, we now strip trailing newlines (found to help / reduce tokens we chew through)
- When reading from OpenAI, we now enforce a pydantic model schema
- When reading from OpenAI, we support converting text spans from the model into integer spans suitable for putting into athena (and complaining about the ones we can't convert)

This is still using the antiquated llama2 setup. A future commit will modernize the NLP used for irae.

### Checklist


cumulus_etl/etl/studies/irae/irae_tasks.py
```diff
@@ -0,0 +1,69 @@
+"""Define tasks for the irae study"""
```
mikix, line 1: This file is worth calling out - because after all the rest of the refactoring here, this file is all that's needed to add a new study - basically just define the system & user prompts, define the output structure, and choose a supported NLP model.

PR #432, opened 2025-08-13
feat(irae): add more models

mikix:
- Drop old llama2 model
- Add gpt-oss-120b, gpt4o, and gpt5 models
- Add llama4-scout model, but it isn't tested yet. The code to use it is fine, but the docker compose bit isn't confirmed to work.

Open issues:
- Confirming llama4 works
- Hardcoded the number of GPUs to 4 - this should be dynamic or configurable. Can figure this out later.

### Checklist


dogversioning (approved):
i look forward to our monthly EC2 cost exploding

PR #435, opened 2025-08-20
feat: have NLP tasks read in DxReports as well as DocRefs

mikix:
This commit adds support for dual-resource tasks and then adds DiagnosticReport to the NLP base task class.

This required some vocabulary alignment, as we used "docref" a lot in places that now can take a docref or a dxreport.
- "note" or "note resource": a DocRef or DxReport resource (dict)
- "note text" or "text": the clinical text inside the note

### Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -102,11 +102,10 @@ jobs:
           docker compose run --rm \
             --volume $DATADIR:/in \
             cumulus-etl \
+            nlp \
             /in/input \
             /in/run-output \
             /in/phi \
-            --export-group nlp-test \
-            --export-timestamp 2024-08-29 \
             --output-format=ndjson \
             --task covid_symptom__nlp_results
 
--- a/cumulus_etl/etl/pipeline.py
+++ b/cumulus_etl/etl/pipeline.py
@@ -120,6 +120,7 @@ async def check_available_resources(
     requested_resources: set[str],
     args: argparse.Namespace,
     is_default_tasks: bool,
+    nlp: bool,
 ) -> set[str]:
     # Here we try to reconcile which resources the user requested and which resources are actually
     # available in the input root.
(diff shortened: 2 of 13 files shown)


.github/workflows/ci.yaml
```diff
@@ -102,11 +102,10 @@ jobs:
           docker compose run --rm \
             --volume $DATADIR:/in \
             cumulus-etl \
+            nlp \
```
mikix, line 105: The fact that this change was needed here but not before, tells me we are regression testing the public release instead of the current branch. We have code to build the branch's docker image... not sure why it's not being used here. I can dig into it later, but just an FYI - it's on my todo list.

cumulus_etl/etl/pipeline.py
```diff
@@ -192,7 +196,7 @@ async def run_pipeline(
             await task.init_check()
 
     # Grab a list of all required resource types for the tasks we are running
-    required_resources = set(t.resource for t in selected_tasks)
+    required_resources = set().union(*(t.get_resource_types() for t in selected_tasks))
```
dogversioning, line 199: ok, i hadn't seen this pattern before, so i had to look it up. can you change the comment above to 'Grab a list of all unique required...'?

cumulus_etl/etl/tasks/nlp_task.py
```diff
@@ -139,59 +139,52 @@ async def init_check(cls) -> None:
...
-                    f"Could not connect to NLP server for DocRef {orig_docref['id']}: {exc}"
-                )
-                self.add_error(orig_docref)
-                continue
-            except pydantic.ValidationError as exc:
-                logging.warning(
-                    f"Could not process answer from NLP server for DocRef {orig_docref['id']}: {exc}"
-                )
-                self.add_error(orig_docref)
+            except Exception as exc:
```
dogversioning, line 168: I'm surprised ruff didn't flag this - do you have the generic exception rule turned off? I think it's fine, i was just expecting a manual comment to disable.

dogversioning (approved):

cumulus_etl/etl/pipeline.py
mikix: I was frustrated that I couldn't find a simpler way to "dynamically combine some sets" - feels like something that should have a util method somewhere, I just couldn't find it.

I can add a comment, yeah.

cumulus_etl/etl/tasks/nlp_task.py
mikix: Bare excepts also have the problem of catching `BaseExceptions` like `SystemExit` and `KeyboardInterrupt`, which usually you don't want to intercept.

But often linting tools don't like the generic version either for some programming purity reasons like "you should know what you're catching". But in the real world, you so often just want to say "hey I don't care what happened, I want to handle it and log it".

PR #436, opened 2025-08-20
feat(irae): update the IRAE system & user prompts

mikix:
### Checklist


Dtphelan1 (approved):
Looks good! We've already talked offline about how future versions of this task will need to reference much more complex PydanticModels, but that's still WIP at the moment.

dogversioning (approved):

PR #443, opened 2025-09-02
nlp: add bedrock support (and allow Azure for gpt-oss and llama4)

mikix:
### Checklist

### DIFF
--- a/cumulus_etl/nlp/openai.py
+++ b/cumulus_etl/nlp/openai.py
@@ -1,7 +1,7 @@
 """Abstraction layer for Hugging Face's inference API"""
 
-import abc
 import os
+from collections.abc import Iterable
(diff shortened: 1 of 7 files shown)


cumulus_etl/nlp/openai.py
```diff
@@ -10,37 +10,62 @@
 from cumulus_etl import errors
 
 
-class OpenAIModel(abc.ABC):
```
mikix, line 13: I'm changing away from an abstract class that is inherited by specific-use models like AzureModel and LocalModel. And switching that into a class that looks for all supported methods (azure, bedrock, local) and finds whichever mode is configured. This better supports stuff like llama4, which can do any of the three modes depending on config.

PR #444, opened 2025-09-05
irae: update prompt and pydantic models, bump min python to 3.11

mikix:
### Checklist


cumulus_etl/etl/tasks/nlp_task.py
```diff
@@ -258,52 +260,74 @@ class BaseOpenAiTaskWithSpans(BaseOpenAiTask):
     1. We need to convert the text spans into integer spans, to avoid PHI hitting Athena.
     2. We need to ensure the pyarrow schema shows a list of ints not strings for spans.
 
-    It assumes the field is named "spans" in the top level of the pydantic model.
+    It assumes any field named "spans" in the hierarchy of the pydantic model should be converted.
```
mikix, line 263: This file's changes are mostly around handling a deeply nested pydantic model with "spans" fields anywhere in the hierarchy (previously, we only needed to handle one-level-deep models, but irae is expanding our needs)

Dtphelan1 (approved):
Pydantic changes look good to me!

PR #449, opened 2025-09-17
nlp: confirm number of documents before running NLP

mikix:
### Checklist


mikix:
> should there be a test of the cli input modes?

Yes probably. :( I was being lazy because I'd have to set up fake ttys :)

PR #456, opened 2025-10-06
nlp: improve bedrock support

mikix:
Now we should actually work for multiple models, using Bedrock's own API, not the OpenAI API (which only works for OpenAI models).

This also adds --provider as an argument for `cumulus-etl nlp`, which specifies which NLP provider to use (azure, bedrock, or local). The default is local, and if you try to use the wrong model, we now warn about it in the console (and I added a model support table to the docs).

Previously we were using environment variables to signal which to use, but now there won't be an obvious difference between bedrock and local, since bedrock just uses your typical AWS config, rather than custom env vars.

### Checklist


cumulus_etl/etl/nlp/cli.py
```diff
@@ -28,6 +28,13 @@ def define_nlp_parser(parser: argparse.ArgumentParser) -> None:
     cli_utils.add_aws(parser, athena=True)
     nlp.add_note_selection(parser)
 
+    parser.add_argument(
+        "--provider",
+        choices=["azure", "bedrock", "local"],
+        default="local",
+        help="which model provider to use (default is local)",
+    )
```
mikix, line 36: Notable changes in this PR, stuffing here:
- This change here, where now users need to specify `--provider` rather than just setting env vars.
- Cache format changed - we no longer cache the direct response from the server, but rather our interpretation of the response from the server.
  - implication 1: you can switch from bedrock to azure and the same note will still be cached (the main benefit of this change, besides the reduction of code complexity)
  - implication 2: if our interpretation meaningfully changes, we'll have to update task_version and ignore existing caches
- Bedrock support required a lot more manual fussing with the response. Their approach to enforcing a JSON schema is real bad. So we have to do some post-processing to catch the stragglers. This means that the black box of NLP is doing slightly different things between Bedrock and Azure. The alternative is not supporting Bedrock at all... But something to maybe test in the future.
- While here, I added a new IRAE task for Claude Sonnet 4.5 for easier testing. We wanted that anyway, so :shrug: I threw it in here.

dogversioning (approved):

PR #457, opened 2025-10-07
Update vllm models

mikix:
### Checklist


compose.yaml
```diff
@@ -178,18 +178,17 @@ services:
     # Needs 80GB of GPU memory. A g6.12xlarge EC2 instance should work.
     # Docs: https://docs.vllm.ai/projects/recipes/en/latest/Llama/Llama4-Scout.html
     extends: common-base
-    image: vllm/vllm-openai:v0.10.0
+    image: vllm/vllm-openai:v0.11.0
```
mikix, line 181: I still couldn't get llama4 to run, but I got closer. With 0.11.0 and downgrading to 4-bit quantization (which is what gpt-oss above uses), I no longer get memory issues. Instead I'm seeing some complaint about the wrong cuda driver version, which might be user-error. But I didn't want to waste too much time on this, so I stopped there. We're closer, but not yet proven to work.

compose.yaml
```diff
@@ -154,7 +154,7 @@ services:
       # If you update anything here that could affect NLP results, consider updating the
       # task_version of any tasks that use this docker.
       - --model=openai/gpt-oss-120b
-      - --revision=bc75b44b8a2a116a0e4c6659bcd1b7969885f423
+      - --revision=b5c939de8f754692c1647ca79fbf85e8c1e70f8a
```
mikix, line 157: Update to the latest version in hugging face, no big changes as far as I can tell from the commits, but they did adjust the prompt templates. I am not bothering to bump any of the task versions using this model, since I just did that in the last (unreleased) commit due to caching changes. So a good time to squeeze this in.

pyproject.toml
```diff
@@ -2,7 +2,7 @@
 name = "cumulus-etl"
 requires-python = ">= 3.11"
 dependencies = [
-    "aiobotocore < 2.22.0",  # FIXME: temp hotfix for dependency version madness - remove later
+    "s3fs[boto3]",  # specify this early to resolve some botocore/aiobotocore dependency madness
```
mikix, line 5: This seems to fix it; suggestion found online. :shrug:

PR #460, opened 2025-10-15
nlp: print consumed tokens after each model run

mikix:
This helps give folks an idea of cost / size of notes.

Example output:
```
 Reading    ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:11
 Finalizing ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:00
 1 written to irae__nlp_claude_sonnet45                             

 Token usage:
  New input tokens:                 715    
  Input tokens read from cache:     11,993 
  Output tokens:                    864    
  Estimated cost (as of Oct 2025):  $0.02  
```

This reports per-task (not one big report at the end for all NLP tasks in a given run), because costs are per-model.

A future improvement could be trying to calculate the cost up front (by tokenizing all the notes ourselves ahead of time, or just doing a quick and dirty "each note is roughly XXX * number of notes" calculation) and including it in the "are you sure?" prompt we do for NLP runs.

### Checklist

### DIFF
--- a/cumulus_etl/etl/tasks/base.py
+++ b/cumulus_etl/etl/tasks/base.py
@@ -155,6 +155,8 @@ class EtlTask:
                 for formatter in self.formatters:
                     formatter.finalize()
 
+        self.finish_task()
+
         return self.summaries
 
     @classmethod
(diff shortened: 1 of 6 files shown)


dogversioning (approved):
i suppose it's too much to ask for prices to be available from an API?

PR #464, opened 2025-10-16
ci: stop pulling newer docker image and use built regression image

mikix:
### Checklist

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -101,6 +101,7 @@ jobs:
 
           # Run the NLP task
           docker compose run --rm \
+            --pull never \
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -101,6 +101,7 @@ jobs:
 
           # Run the NLP task
           docker compose run --rm \
+            --pull never \
```
mikix: This issue got introduced when I added `pull_policy: always` to the compose.yaml file, to force most folks to always get the latest image. But here, it's working against us since I want to use the just-built testing image. So override the pull policy from the CLI.

PR #468, opened 2025-10-23
nlp: be more flexible with the answers accepted from claude

mikix:
And do some light refactoring of NLP classes that will make a future addition of multiple-run prompts easier.

### Checklist

### DIFF
--- a/cumulus_etl/nlp/models.py
+++ b/cumulus_etl/nlp/models.py
@@ -144,8 +144,12 @@ class BedrockProvider(Provider):
             if "toolUse" in content:
                 raw_json = content["toolUse"]["input"]
                 # Sometimes (e.g. with claude sonnet 4.5) we get a wrapper field of "parameter"
-                if len(raw_json) == 1 and "parameter" in raw_json:
-                    raw_json = raw_json["parameter"]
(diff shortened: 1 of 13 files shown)


cumulus_etl/nlp/models.py
```diff
@@ -144,8 +144,12 @@ async def prompt(self, system: str, user: str, schema: type[BaseModel]) -> Promp
...
                 raw_json = content["toolUse"]["input"]
                 # Sometimes (e.g. with claude sonnet 4.5) we get a wrapper field of "parameter"
-                if len(raw_json) == 1 and "parameter" in raw_json:
-                    raw_json = raw_json["parameter"]
+                # or "$PARAMETER_NAME" :shrug: - for now, just look for those names in particular.
+                # If we see a wider variety, we can try to skip any single wrapper field, but I
+                # want to keep the option of studies that have only one top level field for now.
+                top_keys = set(raw_json)
+                if len(top_keys) == 1 and {"parameter", "$PARAMETER_NAME"} & top_keys:
+                    raw_json = raw_json.popitem()[1]
```
mikix: This is the real fix in this PR - everything else was just some light refactoring/updating.

PR #469, opened 2025-10-30
irae: promote naive dates to aware dates before comparing

mikix:
This fixes an error you can get if your notes contain both types.

### Checklist


dogversioning (approved):
image
who brought this guy to the office today

PR #480, opened 2025-12-09
irae: adds serostatus to donor characteristics, updates task version and tests

Dtphelan1:
### Checklist


cumulus_etl/etl/studies/irae/irae_tasks.py
```diff
@@ -28,8 +28,9 @@
 
 
 class BaseIraeTask(tasks.BaseModelTaskWithSpans):
-    task_version = 6
+    task_version = 7
     # Task Version History:
+    # ** 7 (2025-12): Serostatus mentions added to donor task
```
mikix: OK, up to now they've all subtasks shared the same task_version - but that doesn't *need* to be true - we could override it for each of the four(?) subtasks and allow them to drift separately. Since this only invalidates previous results for the donor task, maybe that's worth doing here?

mikix (approved):

PR #534, opened 2026-04-10
Port to cfs.FsPath

mikix:
This was a big change. I had an existing half-baked abstraction (`store.Root` -- originally just for root folders, but grew to include files too, making the name bad. And the API was shit and shared between its own object API and some helpers in the `common` module.). But it wasn't used consistently, and I have been really annoyed at the mix of strings and Root objects.

So I took this opportunity to really commit to "all paths are FsPath objects" throughout. And I wanted to ensure that at the earliest boundary - so argparse arguments that are paths take the flag `type=cfs.FsPath` and things mostly work from there.

This strategy though, turned out to be really annoying for porting tests, which have all sorts of string paths from `tempdir` and the like. But that's the price to pay I guess.

### Checklist


cumulus_etl/etl/config.py
```diff
@@ -129,21 +129,21 @@ def as_json(self):
         }
 
 
-def _latest_config(output_root: store.Root) -> dict:
+def _latest_config(output_root: cfs.FsPath) -> dict:
```
dogversioning, line 132: should this var (and other functions taking this arg) still be called `output_root`, since it's no longer a store.Root? I could see the case for it but for some reason i lightly bumped on this.

I have this kind of comment generally, since i'm seeing it in loaders and elsewhere, but I suspect the answer will be 'its fine', so i'm not going to go too pedantic on this.

cumulus_etl/etl/context.py
```diff
@@ -53,19 +50,19 @@ def last_successful_input_dir(self) -> str | None:
 
     @last_successful_input_dir.setter
     def last_successful_input_dir(self, value: str) -> None:
```
dogversioning: here and below, the type for value must be no longer only str, if you need to cast it to str?

cumulus_etl/nlp/utils.py
```diff
@@ -70,52 +71,42 @@ def get_note_date(note: dict) -> datetime.datetime | None:
     return None
 
 
-def _cache_metadata_filename(cache_dir: str, namespace: str, filename: str) -> str:
-    cache_dir = store.Root(cache_dir, create=True)
-    path = f"nlp-cache/{namespace}/{filename}"
-    return cache_dir.joinpath(path)
+def _cache_metadata_path(cache_dir: str, namespace: str, filename: str) -> cfs.FsPath:
```
dogversioning: cache_dir is of type FsPath, i believe? and also all the functions calling this are also receiving FsPaths? check_metadata_read and check_metadata write are also string typed

cumulus_etl/cli_utils.py
```diff
@@ -126,22 +128,20 @@ def make_export_dir(export_to: str | None = None) -> common.Directory:
             errors.BULK_EXPORT_FOLDER_NOT_LOCAL,
         )
 
-    confirm_dir_is_empty(store.Root(export_to, create=True))
+    export_to.makedirs()
+    confirm_dir_is_empty(export_to)
 
-    return common.RealDirectory(export_to)
+    return common.RealDirectory(str(export_to))
```
dogversioning, line 134: should common.Directory/common.RealDirectory take fsPaths instead of strings?

I guess, more irritatingly: should common more or less have FsPaths instead of strings throughout?

cumulus_etl/common.py
```diff
@@ -108,104 +104,38 @@ def warn_on_missing_resources(found_files: dict[str, str | None], resources: set
...
-        json.dump(data, f, indent=indent)
-
-
 @contextlib.contextmanager
-def read_csv(path: str) -> csv.DictReader:
+def read_csv(path: cfs.FsPath) -> csv.DictReader:
     # Python docs say to use newline="", to support quoted multi-line fields
-    with _atomic_open(path, "r", newline="") as csvfile:
+    with path.open(newline="") as csvfile:
         yield csv.DictReader(csvfile)
```
dogversioning, line 126: should cfs have some sort of method to provide this, similar to cfs.read_multiline_json?

cumulus_etl/store.py
dogversioning, line 1: you came :this: close to deleting this file entirely, which now has a single method - should it live someplace else?

dogversioning (approved):

cumulus_etl/store.py
mikix: Yes probably - can move it to common.

cumulus_etl/common.py
mikix: I dunno - there's a fair amount of convenience I/O API we _could_ add - like this csv reader or NDJSON writing.

I was drawing the line at "would Library use this?" and I left most of the ETL-specific I/O stuff built on top of FsPath here. We could always move it down later, as we need to share it?

cumulus_etl/cli_utils.py
mikix: `common.Directory` is an abstraction of the results of `tempdir.TemporaryDirectory` - so that methods could return either a temp dir that dies when the object does or a real dir that doesn't really have a lifecycle.

In that world, it needs to take a string to look like a TemporaryDirectory. But I suppose I could invert this and use an abstraction that is oriented around FsPath and in the tempdir case, it stuffs a TemporaryDirectory object in a field so that it dies at the right time?

But... I'd be happier to leave that to a later cleanup and suffer a little bit of strings poking through for now.

cumulus_etl/nlp/utils.py
mikix: ah yeah whoops

cumulus_etl/etl/context.py
mikix: yup, whoops. So may strings to remove, not enough time :smile:

cumulus_etl/etl/config.py
mikix: You're not wrong, but I didn't have the energy to change all those lines - I was hoping it would be an easier PR if all those variables weren't changing too.

But you're right that they should probably be `paths` now, not `roots` - but `root` didn't bump me / bother me as a variable, so I figured it was better to have a cleaner PR and clean them up over time as they annoyed me in the code I was working in.

I could get more aggressive about it, but it sounds like you are only half-bothered by it

PR #540, opened 2026-04-16
Adds a float32 pyarrow case in mapping pydantic/json schemas to pyarrow

Dtphelan1:
Seems like we thought we already supported this based on some mocked cli tests, but we didn't! Feels a bit small to add a regression test, but I'll take a stab at one if we think that's necessary just LMK. 

### Checklist

### DIFF
--- a/cumulus_etl/etl/tasks/nlp_task.py
+++ b/cumulus_etl/etl/tasks/nlp_task.py
@@ -386,6 +386,8 @@ class BaseModelTask(BaseNlpTask):
             return pyarrow.bool_()
         elif issubclass(annotation, int):
             return pyarrow.int32()
+        if issubclass(annotation, float):
+            return pyarrow.float32()
         elif issubclass(annotation, enum.Enum):
             return pyarrow.string()  # for now, assume all enums are strings
         elif issubclass(annotation, pydantic.BaseModel):
--- a/cumulus_etl/etl/tasks/nlp_task.py
+++ b/cumulus_etl/etl/tasks/nlp_task.py
@@ -386,7 +386,7 @@ class BaseModelTask(BaseNlpTask):
             return pyarrow.bool_()
         elif issubclass(annotation, int):
             return pyarrow.int32()
-        if issubclass(annotation, float):
+        elif issubclass(annotation, float):
             return pyarrow.float32()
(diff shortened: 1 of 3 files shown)


cumulus_etl/etl/tasks/nlp_task.py
```diff
@@ -386,6 +386,8 @@ def _convert_type_to_pyarrow(cls, annotation) -> pyarrow.DataType:
             return pyarrow.bool_()
         elif issubclass(annotation, int):
             return pyarrow.int32()
+        if issubclass(annotation, float):
```
dogversioning: this should be an elif, and is probably what caused the unit test to break

Dtphelan1: I can't test these unit tests locally still, so sorry I missed that 👎

Dtphelan1:
Conversations offline resulted in: 
- The creation of #541 and #542 after noticing issues with the same tests on main/other branches.
- Temporary updates to deltalake related fixtures to pacify failing tests.

dogversioning (approved):

tests/etl/test_tasks.py
```diff
@@ -340,3 +342,32 @@ def make_formatter(dbname: str, **kwargs):
...
+
+        class Model(pydantic.BaseModel):
+            string_field: str
+            boolean_field: bool
+            integer_field: int
+            float_field: float
+
+        fields = Model.model_fields
+        schema = tasks.BaseModelTask.convert_pydantic_fields_to_pyarrow(fields)
+        print(schema)
```
dogversioning: ✂️

tests/etl/test_tasks.py
```diff
@@ -340,3 +342,32 @@ def make_formatter(dbname: str, **kwargs):
         self.make_json("Device", "A")
         summaries = await basic_tasks.DeviceTask(self.job_config, self.scrubber).run()
         self.assertTrue(summaries[0].had_errors)
+
+
+@ddt.ddt
```
dogversioning: You don't need to include ddt here, since you're not passing multiple configurations into this test.

Dtphelan1: great point!

PR #552, opened 2026-05-07
NLP: Updating GPT5 Model Options

Dtphelan1:
After creating a number of new deployments in Azure, this PR updates the options available to NLP tasks. Main changes: 
- Pivot old GPT5 model (which doesn't exist anymore on Azure, damned fast-pace naming shifts) to GPT5.1 (which we have a batch-endpoint version of)
- Add various mini/nano/default GPT-5.4 options
- Update various tests that used GPT5 to use GPT51

### Checklist


cumulus_etl/nlp/__init__.py
```diff
@@ -5,8 +5,11 @@
...
     Gpt4Model,
     Gpt4oModel,
-    Gpt5Model,
     Gpt35Model,
+    Gpt51Model,
+    Gpt54MiniModel,
+    Gpt54Model,
+    Gpt54NanoModel,
     GptOss120bModel,
     Llama4ScoutModel,
```
Dtphelan1, line 14: Not exactly sure why ruff has this as the preferred sorting order... Happy to update this if people can figure out how to do so while ensuring ruff doesn't yell at me.

dogversioning: It's alphasorted - that's probably right?

mikix: Well, it's alpha-sorted-with-human-numbers (i.e. 5 goes before 35 goes before 51)

dogversioning (approved):

mikix (approved):

PR #555, opened 2026-06-16
Allow explicit resource exclusion

dogversioning:
This supplements `--allow-missing-resources` with a new cli arg, `--exclude-resources`, which allows you to state explicitly that resources should not be present.

It also adds some light fixes for some linux specific behavior, and adjusts one error message to reflect a new arg name.

### Checklist

### DIFF
--- a/cumulus_etl/etl/tasks/task_factory.py
+++ b/cumulus_etl/etl/tasks/task_factory.py
@@ -84,15 +84,18 @@ def get_selected_tasks(
     names: Iterable[str] | None = None,
     *,
     nlp: bool = False,
+    exclusions: Iterable[str] = []
 ) -> list[type[AnyTask]]:
     """
     Returns classes for every selected task.
 
     :param names: an exact list of which tasks to select
     :param nlp: whether we are selecting from NLP or normal tasks
+    :param exclude: a list of resources to explicitly exclude from a task list
     :returns: a list of selected EtlTask subclasses, to instantiate and run
     """
     names = set(cli_utils.expand_comma_list_arg(names, casefold=True))
+    exclusions = set(cli_utils.expand_comma_list_arg(exclusions, casefold=True))
 
     # If we are in NLP mode, we can only select NLP tasks and vice versa.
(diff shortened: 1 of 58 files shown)


cumulus_etl/etl/tasks/task_factory.py
```diff
@@ -84,16 +84,21 @@ def get_selected_tasks(
...
 
     :param names: an exact list of which tasks to select
     :param nlp: whether we are selecting from NLP or normal tasks
+    :param exclude: if true, exclude by, rather than include by, names
     :returns: a list of selected EtlTask subclasses, to instantiate and run
     """
     names = set(cli_utils.expand_comma_list_arg(names, casefold=True))
+    exclusions = set(cli_utils.expand_comma_list_arg(exclusions, casefold=True))
 
+    print(names)
```
Luke-Sikina: What's the difference between `print` and `rich.print`? Should we exclusively be using one or is there a reason for the mixture?

cumulus_etl/etl/tasks/task_factory.py
```diff
@@ -84,15 +84,18 @@ def get_selected_tasks(
...
     *,
     nlp: bool = False,
+    exclusions: Iterable[str] = []
 ) -> list[type[AnyTask]]:
     """
     Returns classes for every selected task.
 
     :param names: an exact list of which tasks to select
     :param nlp: whether we are selecting from NLP or normal tasks
+    :param exclude: a list of resources to explicitly exclude from a task list
```
Luke-Sikina: This doc seems mismatched with the function signature. I see `exclude: bool` here, but `exclusions: Iterable[str]` on 87.

Luke-Sikina (approved):

cumulus_etl/etl/tasks/task_factory.py
dogversioning: bare `print` is almost certainly debug - i have a pre-commit hook on some other repos to catch this since it is a sin i constantly commit, i'll add one here as well.

PR #556, opened 2026-06-16
Cutover CI to `uv`

dogversioning:
For speed reasons, and to workaround aioboto version churn, we'll use `uv` for package installation in ci jobs.

### Checklist


Dtphelan1 (approved):
Looks fine to me!

Issue #559, opened 2026-07-08
Upload Notes: Should be able to upload up until a cap is hit


Dtphelan1:
When uploading notes to a LabelStudio project, it would be nice to set a cap on thbe number of notes I want to upload to a project. Additionally, it would be nice if etl created an artifacts (even just to stdout) that told me which notes were uploaded by note_ref or ID. 

Motivation: Since some aspects of NLP are stochastic and our sample-selection criteria are imperfect, there is no guarantee that the original sample create using a study will all have  NLP annotations for downstream tasks. The result is that often we oversample in the early stages (e.g. run NLP on 3k notes) in order to ensure we have sufficient samples for later stages (e.g. 300 notes to run agreement processes on). Since the final upload-notes stage is usually targetting specific `N` for statisical reasons, it would be helpful to be able to CLI specify how many notes I want uploaded to a project.

PR #560, opened 2026-07-20
`upload-notes` now exposes arguments to limit uploads by chart count

Dtphelan1:
The `upload-notes` command now takes some new arguments to enable user-defined limits on how many charts are uploaded to a labelstudio project, closing #559. New Arguments include: 
- `count`: How many charts (after grouping, if requested) should be uploaded to the labelstudio instance
- `seed`: Random seed used in sampling, to provide deterministic sampling when desired.

Docs are updated accordingly. I also introduce the idea of creating a manifest after note-selection is done if there is an `--export-to` directory specified. This can be a helpful artifact for replicability, and any PHI concerns should be side-stepped by the dir will already have exported patient PHI in it. 

If you notice the weird irrelevant file changes, it's because I ran `ruff format`. Shocked that these changes didn't impact CI, but presumably the formatting isn't mandated. If you feel strongly we can roll them back. 

### Checklist

### DIFF
--- a/cumulus_etl/etl/cli.py
+++ b/cumulus_etl/etl/cli.py
@@ -131,7 +131,7 @@ async def etl_main(args: argparse.Namespace) -> None:
         deleted_ids=loader_results.deleted_ids,
         export_group_name=export_group,
         export_datetime=export_datetime,
-        exclusions=','.join(args.exclusions),
+        exclusions=",".join(args.exclusions),
         format_kwargs={"optimize_table": args.table_optimization},
     )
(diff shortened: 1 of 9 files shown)


dogversioning (approved):
> If you notice the weird irrelevant file changes, it's because I ran ruff format. Shocked that these changes didn't impact CI, but presumably the formatting isn't mandated. If you feel strongly we can roll them back.

Interesting - do you have a different local version, maybe? do you have the pre-commit hooks installed? maybe the CI should do a format check if it's not already?

Issue #548, opened 2026-04-28
Ugly traceback error if auth error happens during `init`


mikix:
To reproduce:

```
cumulus-etl init s3://bucket/folder/
```

without any AWS_PROFILE or anything. You'll get a gross auth traceback. But it should be a nicer, user-friendly error, suggesting to set AWS_PROFILE.

I think it also happens if the wrong --s3-kms-key is provided.

PR #561, opened 2026-07-24
Add a specific try/except block for AWS Authentication issues.

msa2984:
Addresses #548 

### Checklist

Notes:
- We could pull in a more direct import for referencing the class (vs a hardcoded string) from the jvm package, but this seemed to add a lot of code in comparison to the desired functionality. 
- In each test I ran, it seemed more prudent to handle the more specific "NoAuthWithAWS" exception than the AccessDenied exception because AccessDenied for files can be caused by a multitude of issues, not just the traceback indicated in the issue being addressed. NoAuthWithAWS is always nested within this parent exception.

Test:
- New functionality:
  - Ensure the `AWS_PROFILE` IS NOT set in your development environment / Credentials are not available.
  - Run `init` .. An example:
  ```sh
  docker compose run --build \
  cumulus-etl init \
  s3://your_bucket_name
   ```
  - Verify the error: `Could not access the provided S3 bucket. Verify AWS_PROFILE is configured before execution.` displays in the terminal, and no folders + files were added to the S3 bucket.
- Typical functionality: 
   - Ensure the `AWS_PROFILE` IS set in your development environment / Credentials are available.
   - Run `init` .. An example:
  ```sh
  docker compose run --build \
  cumulus-etl init \
  s3://your_bucket_name
   ```
  - Verify the folder structure and related files have been created in the S3 bucket.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,8 +31,12 @@ This will install the pre-commit hooks for the repo (which automatically enforce
 
 1. First, you'll want to install a Java JDK, for Delta Lake support.
 
-2. Then just run `pytest`.
-All dependencies should have been installed by the `pip install .[dev]` above.
+2. To ensure all test dependencies are installed, run this command:
+```sh
+pip install .[tests]
+```
+
+3. Then just run `pytest`.
 
 ### How to show us the patch
 
--- a/cumulus_etl/formats/base.py
+++ b/cumulus_etl/formats/base.py
@@ -7,6 +7,7 @@ from collections.abc import Collection
 import cumulus_fhir_support as cfs
 
 from cumulus_etl.formats.batch import Batch
+from py4j.protocol import Py4JJavaError
 
 
 class Format(abc.ABC):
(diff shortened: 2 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -31,8 +31,12 @@ This will install the pre-commit hooks for the repo (which automatically enforce
 
 1. First, you'll want to install a Java JDK, for Delta Lake support.
 
-2. Then just run `pytest`.
```
msa2984, line 34: I can move this to a separate PR, however these instructions seem to be out-of-date as pytest is marked as an optional-dependency and so tests did not run successfully until I also installed .[tests]

cumulus_etl/formats/base.py
```diff
@@ -66,9 +69,24 @@ def write_records(self, batch: Batch) -> bool:
...
             self._write_one_batch(batch)
             return True
+        except Py4JJavaError as e:
+            class_name = e.java_exception.getClass().getName()
+
+            nested_exception = e.java_exception.getCause()
+            if nested_exception:
+                class_name = nested_exception.getClass().getName()
+
+            if class_name == AWS_AUTH_EXCEPTION_CLASS_NAME:
```
Luke-Sikina, line 79: Nit: not a huge fan of pulling out constants in this context. It's used in 1 place. It's unlikely to change. It means I have to do a little more work to understand the code. I'd rather just have the string here.

msa2984: yes but pulling it out into a constant means that we can use the constant in the corresponding unit test (although I don't mind changing it back)

tests/etl/test_tasks.py
```diff
@@ -203,6 +205,30 @@ async def test_deleted_ids(self):
 class TestTaskCompletion(TaskTestCase):
     """Tests for etl__completion* handling"""
 
+    async def _configure_completion_write_task(
```
msa2984, line 208: this is a DRY preference - I can move this back into its respective tests,  `test_error_during_write`, `test_py4jjavaerror_during_write`, and `test_noauthwithawsexception_during_write`

cumulus_etl/formats/base.py
```diff
@@ -5,9 +5,12 @@
 from collections.abc import Collection
 
 import cumulus_fhir_support as cfs
+from py4j.protocol import Py4JJavaError
```
dogversioning: nit, here and elsewhere: we've got a weak preference for module level imports, so `from py4j import protocol`

cumulus_etl/formats/base.py
dogversioning: I tend to be on the side of 'use the constant', myself.

dogversioning (approved):

PR #566, opened 2026-09-14
`upload-notes` Argument Extension: `--export-labels-to` and `--no-upload` for Label Generation

Dtphelan1:
This PR adds support for generating label files for use by chart-review, detailing the annotations for a given LLM. This is critical for comparing agreement between models and other annotators (e.g. clinical chart-reviewers themselves). In lieu of a separate command, I've bolted on these as arguments to upload-notes. The `--export-labels-to` argument seems very useful in general for exporting at upload-note runtime, which will be the typical use-case. `--no-upload`'s utility will be largely ad-hoc, helpfully generating labels for notes that have already been uploaded. 

You may be curious as to why we generate separate doc_ref and diagnostic_report label files. AFAICT way that [chart-review](https://github.com/smart-on-fhir/chart-review/blob/5da4445bacfb9b44d1f9f1c7a0bd8ebc0c9c1d41/chart_review/external.py#L40-L78) works today assumes a single column view based on resource type. This implementation aims to align with those downstream expectations. Future updates to chart-review though might look into a more unified conceptualization of charts/chart-ids, that allows for multiple types of things (docref and dxr, for example) to all be equivalent `charts` for a given comparison. I'll make an issue to this effect.

Docs are also updated

### Checklist


cumulus_etl/upload_notes/cli.py
```diff
@@ -379,6 +411,13 @@ def define_upload_notes_parser(parser: argparse.ArgumentParser) -> None:
         help="name of an Athena table with annotations "
         "(must have note ID, label, and span columns)",
     )
+    group.add_argument(
+        "--export-labels-to",
+        metavar="PATH",
+        type=cfs.FsPath,
+        help="where to write Chart Review label files for the uploaded notes "
```
msa2984, line 418:
```suggestion
        help="where to write Chart Review label files for the uploaded notes"
```

cumulus_etl/upload_notes/cli.py
```diff
@@ -392,19 +431,23 @@ def define_upload_notes_parser(parser: argparse.ArgumentParser) -> None:
...
         help="Label Studio project ID to update",
-        required=True,
     )
     group.add_argument(
         "--overwrite", action="store_true", help="whether to overwrite an existing task for a note"
     )
+    group.add_argument(
+        "--no-upload",
+        action="store_true",
+        help="prepare the notes but don't push them to Label Studio "
```
msa2984, line 448:
```suggestion
        help="prepare the notes but don't push them to Label Studio"
```

docs/chart-review.md
```diff
@@ -176,6 +179,67 @@ These all expect a certain format:
...
+ cumulus-etl upload-notes \
+  --ls-token /host/label-studio-token.txt \
+  --ls-project 3 \
+  --label-by-athena-table my_study__nlp_labels \
+  --export-labels-to /host/chart-review-project/ \
+  /host/my-input-folder/ \
+  https://my-label-studio-server/ \
+  s3://my-cumulus-prefix-phi-99999999999-us-east-2/subdir/
+```
+
```
msa2984, line 200: Mostly a nit because this doc is great, you might want to also add a note about `--export-to` in here.

cumulus_etl/upload_notes/labels.py
```diff
@@ -0,0 +1,179 @@
...
+def _write_one_file(path: cfs.FsPath, res_type: str, rows: Iterable[tuple]) -> None:
+    rows = list(rows)
+    columns = [ID_COLUMNS[res_type], *LABEL_COLUMNS]
+    # Chart Review errors out on a sublabel_name column without a sublabel_value column, and empty
+    # sublabel columns are just noise for origins that don't use them - so only write the pair
+    # when this origin actually has sublabels.
+    has_sublabels = any(row[2] or row[3] for row in rows)
+    if has_sublabels:
+        columns += SUBLABEL_COLUMNS
+
```
msa2984, line 166: q: is it useful to have a warning here that a column is being skipped?

msa2984 (approved):
A few Qs but nothing blocking from my end.

docs/chart-review.md
Dtphelan1: What a wonderful suggestion!

cumulus_etl/upload_notes/labels.py
Dtphelan1: Feels like a great idea - will add that as well.

cumulus_etl/upload_notes/cli.py
```diff
@@ -30,6 +30,37 @@ def init_checks(args: argparse.Namespace):
         )
 
 
+def check_upload_args(args: argparse.Namespace) -> None:
+    """
+    Confirms we have what we need for the work the user asked for.
+
+    The Label Studio arguments aren't required by argparse itself, because --no-upload runs don't
+    need them at all. So we check them here instead, once we know which kind of run this is.
```
dogversioning: i bounced off this a little bit - how is this different than `init_checks()`, given they're called in order? it could just be at the start of the init function and behave the same?

cumulus_etl/upload_notes/cli.py
```diff
@@ -292,10 +323,11 @@ async def push_to_label_studio(
 
 
 def define_upload_notes_parser(parser: argparse.ArgumentParser) -> None:
-    parser.usage = "cumulus-etl upload-notes [OPTION]... INPUT LS_URL PHI"
+    parser.usage = "cumulus-etl upload-notes [OPTION]... INPUT [LS_URL] PHI"
 
     parser.add_argument("dir_input", metavar="/path/to/input", type=cfs.FsPath)
-    parser.add_argument("label_studio_url", metavar="https://example.com/labelstudio")
+    # Optional so that --no-upload runs don't have to name a server they'll never talk to.
+    parser.add_argument("label_studio_url", metavar="https://example.com/labelstudio", nargs="?")
```
dogversioning: is there a reason you want to not take this as an explicit `--label_studio_url` arg? if so, fine, but if it's optional as you say, i think this is a little more natural/less prone to breakage if you add another non-named param option.

dogversioning (approved):
just some light cli ux nits, appreciate the amount of inline documentation

cumulus_etl/upload_notes/labels.py
Dtphelan1: Changed things around a bit to support this!

cumulus_etl/upload_notes/cli.py
Dtphelan1: Great feedback. WE discussed and decided to go with rearranging the arguments s.t. this is now a named arg. This will lose us backwards compatability with previous docs of this, but that's a small blast-radius to address.

cumulus_etl/upload_notes/cli.py
Dtphelan1: I agree with the logic here of comlbining these, but has a minor diagreement on the ideal naming: I've actually gotten rid of init_checks as as separate method, and nested it under the more generic arg check. Fundamentally accomplishes the same thing, but noting for posterity.

